PR #23, opened 2018-02-01
Adds generation of source maps with debug run

sondreb:
- Adds dependencies gulp-if and gulp-sourcemaps.
- Adds a new gulp task to set environment variable to generate source maps.


sondreb:
Before the changes:

image

After adding source map output:

image

stevshan:
Great! Nice work, @sondreb. The PR is accepted but please notice that there is a modularization work ongoing which refactor these codes to provide a better development integration.  So the codes in the PR may be rewritten later.

PR #50, opened 2018-04-27
multi fixes for accessiblity issues

tonglingchen:
- fix several issues for Windows high contrast mode
- enable keyboard control (left arrow and right arrow) for splitter
- fix several issues with screen reader
- upgrade dependencies


src/Sfx/package.json
```diff
@@ -39,6 +39,8 @@
     "yargs": "7.0.1"
   },
   "dependencies": {
+    "hawk": "^7.0.7",
+    "hoek": "^5.0.3",
```
stevshan: Please also update the third party notice in standalone as the these dependencies will be included in the standalone release.

PR #52, opened 2018-04-30
set min value for parameters on create service dialog

tonglingchen:
Did not set max value as int.max is not user friendly number for screen reader (too big)

### DIFF
--- a/src/Sfx/App/partials/create-service-dialog.html
+++ b/src/Sfx/App/partials/create-service-dialog.html
@@ -54,7 +54,7 @@
             </dd>
             <dt ng-if="!action.serviceType.raw.ServiceTypeDescription.IsStateful">Instance Count</dt>
             <dd ng-if="!action.serviceType.raw.ServiceTypeDescription.IsStateful">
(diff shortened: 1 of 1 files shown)


src/Sfx/App/partials/create-service-dialog.html
```diff
@@ -54,34 +54,34 @@ <h4 class="modal-title">Create Service</h4>
             </dd>
             <dt ng-if="!action.serviceType.raw.ServiceTypeDescription.IsStateful">Instance Count</dt>
             <dd ng-if="!action.serviceType.raw.ServiceTypeDescription.IsStateful">
-                <input type="number" class="input-flat" autocomplete="off" ng-model="action.description.raw.InstanceCount" aria-label="Instance Count">
+                <input type="number" class="input-flat" autocomplete="off" ng-model="action.description.raw.InstanceCount" aria-label="Instance Count" min="1">
```
tonglingchen: I am using min="1" here since this is for Stateful service only.

PR #54, opened 2018-05-05
Events API integration into SFX UI

ibabou:
This mainly to collect comments and feedback from others on code. We probably will wait and work on this in the forked repo till just before 6.3 release, where we will merge this into main SFX repo and include updated package in the new runtime.

Some of the changes I made after the meeting we had (Thanks for all feedback!):
- Adding the events tabs to all entity levels supporting all existing APIs.
- Fixed date-picker control and selection of start and end dates.
- UI fixes in detail list itself including margin for first col if table has expandable rows, fixed widths for primary cols, and having the table flexible to fill window size.
- Changed the position of correlated events link & updated date-time format.
- Refresh after initial load now just retrieves latest hour which should cover any correlated related updates.
- Fixed the format of details, and now second row fills the whole space.
- Handling reset of loading calls in case new dates got set.
- Handling the case where APIs where being used with versions <6.2. Pending: handle 6.2+ where data isn't available.
- Having colors for health events (temporary).


src/Sfx/App/Scripts/Models/DataModels/Collections.ts
```diff
@@ -61,27 +115,54 @@ module Sfx {
...
-                    return this;
-                }).finally(() => {
-                    this.refreshingPromise = null;
-                });
+                this.refreshingPromise =
+                    this.refreshingLoadPromise.load(() => {
+                        return this.retrieveNewCollection(messageHandler);
+                    }).catch((error) => {
+                        if (!error || error.isCanceled !== true) {
+                            throw error;
```
tonglingchen: Not sure if this is by intention - throw error when !error is true?

src/Sfx/App/Scripts/Models/Events.ts
```diff
@@ -0,0 +1,191 @@
...
+            for (const property in responseItem) {
+                if (!this.extractField(property, responseItem[property])) {
+                    this.eventProperties[property] = responseItem[property];
+                }
+            }
+        }
+
+        protected extractField(name: string, value: any): boolean {
+            if (name === "Kind") {
+                this._kind = value;
```
tonglingchen: Why not use switch-case here?

src/Sfx/App/partials/date-picker.html
```diff
@@ -0,0 +1,14 @@
+<div class="col-md-3">
+    <p class="input-group">
+        <input type="text" class="input-flat clear-disabled" uib-datepicker-popup="{{format}}"
+            ng-model="ngModel" is-open="opened"
+            min-date="minDate" max-date="maxDate" init-date="initDate"
+            datepicker-options="dateOptions" placeholder="{{placeHolderText}}" show-button-bar="false"
+            ng-required ng-readonly="readOnly" ng-click="ctrl.openPopup($event)" />
+        <span class="input-group-btn">
+            <a class="calendar-link" ng-click="ctrl.openPopup($event)" >
```
tonglingchen: Please add role="button" as it works like a button instead of a link. 
Also, if clicking on it will open a popup, please set aria-haspopup="true"
Please set href="" so that it can capture the focus

src/Sfx/App/partials/events-view.html
```diff
@@ -0,0 +1,18 @@
...
+        <div class="row" style="margin-right:0px;">
+                <sfx-date-picker ng-model="dwSelector.startDate" min-date="dwSelector.startDateMin"
+                    max-date="dwSelector.startDateMax" init-date="dwSelector.startDateInit" place-holder-text="'Start Date'">
+                </sfx-date-picker>
+                <sfx-date-picker ng-model="dwSelector.endDate" min-date="dwSelector.endDateMin"
+                    max-date="dwSelector.endDateMax" init-date="dwSelector.endDateInit" place-holder-text="'End Date'">
+                </sfx-date-picker>
+                <div class="col-md-3">
+                    <div class="btn-group" style="margin-bottom:3px;">
+                        <a class="reset-link" href="" ng-click="resetClick()">Reset</a>
```
tonglingchen: Please add role="button" as it works like a button instead of a link (which redirects users to either another page, or a different location on the same page)

src/Sfx/App/Scripts/Common/RestClient.ts
```diff
@@ -13,6 +13,7 @@ module Sfx {
         private static defaultApiVersion: string = "3.0";
         private static apiVersion40: string = "4.0";
         private static apiVersion60: string = "6.0";
+        private static apiVersion62preview: string = "6.2-preview";
```
tonglingchen: apiVersion62Preview with upper case P? :smile:

src/Sfx/App/Scripts/Models/DataModels/Collections.ts
ibabou: I just wanted to handle here the case where a different error other than cancellation happens. I believe in this case the passed object will always have a value, so I'll be changing this to error && error.isCanceled !== true.

ibabou:
As discussed, I fixed the font issue. Also, fixed the datepicker to open on focus, and now it can be navigated with the keyboard.

src/Sfx/App/partials/date-picker.html
ibabou: Sure, will do.

src/Sfx/App/partials/events-view.html
ibabou: Sure, will do.

src/Sfx/App/partials/date-picker.html
```diff
@@ -0,0 +1,14 @@
+<div class="col-md-3">
+    <p class="input-group">
+        <input type="text" class="input-flat clear-disabled" uib-datepicker-popup="{{format}}"
+            ng-model="ngModel" is-open="opened"
+            min-date="minDate" max-date="maxDate" init-date="initDate"
+            datepicker-options="dateOptions" placeholder="{{placeHolderText}}" show-button-bar="false"
+            ng-required ng-readonly="readOnly" ng-click="ctrl.openPopup($event)" />
```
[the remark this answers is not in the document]
ibabou: Sure, will do.

src/Sfx/App/Scripts/Models/Events.ts
ibabou: Will change it to switch.

src/Sfx/App/Scripts/Common/RestClient.ts
ibabou: Sure :)

PR #58, opened 2018-05-29
The remoting objects infrastructure of moodularization for SFX Standalone

stevshan:
* Refine the modularization interfaces.
* New design of remoting mechanism.


stevshan:
Need to refresh the commits for the PR.

PR #66, opened 2018-06-15
retire bower

tonglingchen:


src/Sfx-Standalone/.build/tasks/pack.licensing.js
```diff
@@ -283,6 +283,7 @@ function generateThirdPartyNotice(deps, noticeFilePath) {
         fs.appendFileSync(noticeFd, "This project incorporates components from the projects listed below. The original copyright notices and the licenses under which Microsoft received such components are set forth below. Microsoft reserves all rights not expressly granted herein, whether by implication, estoppel or otherwise.\r\n");
         fs.appendFileSync(noticeFd, "\r\n");
 
+        deps = deps.filter((value, index, self) => self.findIndex((item) => item.name === value.name) === index);
```
stevshan: Seems the self and deps share the same index order. so why it's necessary to do a findIndex in the self?

src/Sfx-Standalone/.build/tasks/pack.licensing.js
```diff
@@ -296,6 +297,9 @@ function generateThirdPartyNotice(deps, noticeFilePath) {
             fs.appendFileSync(noticeFd, "=========================================\r\n");
             fs.appendFileSync(noticeFd, utils.format("END OF {} NOTICES AND INFORMATION\r\n", dep.name));
         }
+    }
+    catch (e) {
```
stevshan: It would be better if write some thing about what's the current operation before log the error as all the build tasks are executed paralleled

src/Sfx/gulpfile.js
```diff
@@ -7,6 +7,7 @@ var plugins = require("gulp-load-plugins")({
     camelize: true,
     lazy: true
 });
+var libConfiguration = require("./build/lib.json");
```
stevshan: please use const instead. Can we also change other "var" to const if necessary?

src/Sfx/package.json
```diff
@@ -27,19 +24,31 @@
...
+    "angular-aria": "1.5.0",
+    "angular-route": "1.5.0",
+    "angular-sanitize": "1.5.0",
+    "angular-ui-bootstrap": "2.5.6",
+    "bootstrap-sass": "3.3.7",
+    "d3": "3.5.17",
+    "jquery": "^2.2.3",
+    "jquery-ui": "^1.12.1",
+    "lodash": "^4.17.10",
+    "moment": "^2.8.3",
```
stevshan: Why these become the dependencies of SFX Standalone?

src/Sfx-Standalone/.build/tasks/pack.licensing.js
tonglingchen: This is to remove duplication. Otherwise there will be two angulars and two jquerys.

src/Sfx-Standalone/.build/tasks/pack.licensing.js
tonglingchen: sure.

src/Sfx/package.json
tonglingchen: They are not dependencies of SFX Standalone. However the dependency list in 3rd party notice includes both dependencies from SFX and SFX Standalone.

src/Sfx-Standalone/.build/tasks/pack.licensing.js
stevshan: Set can be used here to reduce the complexity and improve the performance. :)

PR #78, opened 2018-06-21
Package Manager

stevshan:
* Refactor http module.
* Add package manager.
* Hook up the module manager with package manager.
* Make settings module monitor the settings objects.


tonglingchen:
>         getComponentAsync(componentIdentity: "http.https-client"): Promise<IHttpClient>;

not sure why do we need two http-client here..

---
Refers to: src/Sfx-Standalone/modules/http.d.ts:11 in 2d627be. [](commit_id = 2d627bec82e35101761d6088789f6fd28b88c96b, deletion_comment = False)

PR #88, opened 2018-07-03
IPC communicator fix for IpcRenderer

stevshan:
* Decouple ChannelProxy.
* Refactor ChannelProxy.
* Add support of IpcRenderer.send()/sendTo()/sendHost().
* Auto-detect the IpcRenderer scenarios.


src/Sfx-Standalone/modules/ipc/proxy/electron-ipc-renderer.ts
```diff
@@ -0,0 +1,87 @@
...
+            this.channel = undefined;
+        }
+    }
+
+    public sendMessage(msg: IMessage): boolean {
+        if (this.disposed) {
+            throw new Error("Channel proxy already disposed.");
+        }
+
+        if (this.windowId === -2) {
```
tonglingchen: >-2 [](start = 30, length = 2)

better to use const to avoid magical number?

PR #94, opened 2018-07-13
Change the installer from perMachine to perUser.

stevshan:
* Add new files.msi.wxs.xslt to update the files.msi.wxs for registry/uninstall.
* Add publish.windows.update-guid.ps1 to update Id/Guid attribute in files.msi.wxs.
* Suppress ICE91 ICE60 warnings for WIX as it's intended to install only for the current user.

### DIFF
--- a/src/Sfx-Standalone/.build/files.msi.wxs.xslt
+++ b/src/Sfx-Standalone/.build/files.msi.wxs.xslt
@@ -1,0 +1,148 @@
+<?xml version="1.0" encoding="utf-8"?>
+<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+    xmlns:msxsl="urn:schemas-microsoft-com:xslt" exclude-result-prefixes="msxsl"
+    xmlns:wix="http://schemas.microsoft.com/wix/2006/wi">
+  <xsl:output method="xml" indent="yes"/>
+  <xsl:template match="@*|node()">
+    <xsl:copy>
+      <xsl:apply-templates select="@*|node()"/>
+    </xsl:copy>
+  </xsl:template>
+
+  <xsl:template match='/wix:Wix/wix:Fragment/wix:ComponentGroup/wix:Component/wix:File'>
+    <xsl:copy>
+      <xsl:apply-templates select="@*"/>
+      <xsl:attribute name="KeyPath">
+        <xsl:text>no</xsl:text>
+      </xsl:attribute>
+    </xsl:copy>
+  </xsl:template>
+
+  <xsl:template match="/wix:Wix/wix:Fragment/wix:ComponentGroup">
+    <xsl:copy>
+      <xsl:apply-templates select="@*"/>
+
+      <xsl:element name="Component" namespace="http://schemas.microsoft.com/wix/2006/wi">
+        <xsl:attribute name="Id">
+          <xsl:text>*</xsl:text>
+        </xsl:attribute>
+        <xsl:attribute name="Directory">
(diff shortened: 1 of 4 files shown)


src/Sfx-Standalone/.build/files.msi.wxs.xslt
```diff
@@ -0,0 +1,148 @@
+<?xml version="1.0" encoding="utf-8"?>
```
tonglingchen: this is crazy... we have to do all of these to get things work?

src/Sfx-Standalone/.build/tasks/publish.windows.js
```diff
@@ -56,23 +56,26 @@ gulp.task("publish:msi@build-msi",
...
         const candlePath = path.resolve("./.vendor/wix/candle.exe");
         const lightPath = path.resolve("./.vendor/wix/light.exe");
-        const heatCmd = utils.format("\"{}\" dir \"{}\" -ag -srd -cg MainComponentsGroup -dr INSTALLFOLDER -o \"{}\"", heatPath, packDirPath, filesWixPath);
+        const heatCmd = utils.format("\"{}\" dir \"{}\" -ag -srd -cg MainComponentsGroup -dr INSTALLFOLDER -o \"{}\" -t \"{}\"", heatPath, packDirPath, filesWixPath, filesWixXsltPath);
+        const updateGuidCmd = utils.format("powershell \"{}\" -XmlPath \"{}\"", path.resolve("./.build/tasks/publish.windows.update-guid.ps1"), filesWixPath);
         const candleCmd = utils.format("\"{}\" -arch x86 -out \"{}\\\\\" \"{}\" \"{}\"", candlePath, wxsobjDir, path.resolve(path.join(buildInfos.paths.buildDir, "msi.wxs")), filesWixPath);
         const lightCmd =
             utils.format(
-                "\"{}\" -b \"{}\" -spdb -out \"{}\" \"{}\" \"{}\"",
+                "\"{}\" -b \"{}\" -spdb -out \"{}\" \"{}\" \"{}\" -sice:ICE60 -sice:ICE91",
```
tonglingchen: ICE60 means files won't be uninstalled after user uninstall the MSI, right? Is it possible to fix them by adding <removefolder> for each folder in xslt?

stevshan: No, ICE60 means the font file needs a language specification but as SFX is not multi-language tool. 
https://docs.microsoft.com/en-us/windows/desktop/msi/ice60

src/Sfx-Standalone/.build/files.msi.wxs.xslt
stevshan: Well, WIX not a active project anymore and SFX is a ongoing project. So there may be more requirements coming. It would be better if we follow the pattern now before the code become too big to refactor.

src/Sfx-Standalone/.build/tasks/publish.windows.js
tonglingchen: Got it. thanks.

src/Sfx-Standalone/.build/files.msi.wxs.xslt
tonglingchen: What pattern? Just saying there should be a better way to do packing/installation than WIX :)

stevshan: I mean the WIX pattern. Yeah, we may need to find a better packing way on Windows.

stevshan: Just gave WIX a second look. Seems WIX is now back to active again.

PR #111, opened 2018-08-30
Update error messages if eventstore system service is not enabled

ibabou:


src/Sfx/App/Scripts/Common/ResponseMessageHandlers.ts
```diff
@@ -75,12 +75,16 @@ module Sfx {
         public getErrorMessage(apiDesc: string, response: ng.IHttpPromiseCallbackArg<any>): string {
             // API is not available on this cluster.
             if (response.status === 400) {
-                return "EventsStore is not available on current cluster.";
+                return "Events API is not available on current cluster.";
+            }
+            // EventStoreService system service is not added (>=6.4).
+            if (response.status === 404 && response.statusText === "FABRIC_E_SERVICE_DOES_NOT_EXIST") {
+                return "EventStore system service is not found, please enable it.";
```
tonglingchen: I think it would be better to show this error message in the event list result area. Right now this error message will show for just 2 or 3 seconds and then gone. Something we can think about next release.

ibabou: I agree. I'm not sure if you have a control in place where I can set messages and so it is shown there. I think it'll be better, this bottom one is used more to show transient errors/failures.

PR #114, opened 2018-09-12
Fix the logics to cache the members of the object proxy.

stevshan:
* Only cache the non-writable, non-configurable, no get/set accessors members.
* Only cache the non-enumerable and function members.

### DIFF
--- a/src/Sfx-Standalone/modules/proxy.object/data-info-manager.ts
+++ b/src/Sfx-Standalone/modules/proxy.object/data-info-manager.ts
@@ -153,7 +153,7 @@ export class DataInfoManager implements IDisposable {
 
     private toObjectDataInfo(target: Object, parentId?: string): IDataInfo {
         const ref = this.refRoot.refer(target, parentId);
-        
+
         let dataInfo: IObjectDataInfo = <IObjectDataInfo>ref.getRefDataInfo(target);
 
         if (dataInfo) {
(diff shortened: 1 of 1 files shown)


src/Sfx-Standalone/modules/proxy.object/data-info-manager.ts
```diff
@@ -172,14 +172,18 @@ export class DataInfoManager implements IDisposable {
...
         while (currentObj && currentObj !== Object.prototype) {
             const propertyDescriptors = Object.getOwnPropertyDescriptors(currentObj);
+            const isClass = Function.isFunction(propertyDescriptors["constructor"]);
 
             for (const propertyName in propertyDescriptors) {
                 const propertyDescriptor = propertyDescriptors[propertyName];
 
-                if (!propertyDescriptor.enumerable
-                    || !propertyDescriptor.writable
+                if ((!propertyDescriptor.writable
```
tonglingchen: Can you add comments to describe these condition checks.

src/Sfx-Standalone/modules/proxy.object/data-info-manager.ts
```diff
@@ -172,14 +172,18 @@ export class DataInfoManager implements IDisposable {
...
+                    && !propertyDescriptor.configurable
                     && !propertyDescriptor.get
-                    && !propertyDescriptor.set) {
+                    && !propertyDescriptor.set)
+                    || (isClass 
+                        && !propertyDescriptor.enumerable
+                        && Function.isFunction(propertyDescriptor.value))) {
```
tonglingchen: what is the purpose of following check ? 
Function.isFunction(propertyDescriptor.value)

stevshan: For members of a class, only cache those which are not enumerable functions, whose values are more stable.

PR #121, opened 2018-09-18
Fix the logics to check if the object is serializable.

stevshan:
* If the value is an array, check every items instead.
* If the value is an object, check every enumerable properties.

### DIFF
--- a/src/Sfx-Standalone/utilities/utils.ts
+++ b/src/Sfx-Standalone/utilities/utils.ts
@@ -70,13 +70,15 @@ Object.isSerializable = (value: any) => {
                 return value[Symbols.Serializable] === true;
             }
 
-            if (Array.isArray(value)) {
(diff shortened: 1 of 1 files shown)


src/Sfx-Standalone/utilities/utils.ts
```diff
@@ -70,9 +70,13 @@ Object.isSerializable = (value: any) => {
                 return value[Symbols.Serializable] === true;
             }
 
+            if (Array.isArray(value)) {
```
tonglingchen:
```javascript
if (Function.isFuction(value["toJSON"]) {
    return true;
}

if (Array.isArray(value)) {
    return value.every((itemValue) => Object.isSerializable(itemValue));
}

return Object.values(value).every((propertyValue) => Object.isSerializable(propertyValue));

PR #146, opened 2018-11-06
restart replica/deployed replica

jeffj6123:
Added the ability to use restart replica/deployed replica as actions.

### DIFF
--- a/src/Sfx/App/Scripts/Services/RestClientService.ts
+++ b/src/Sfx/App/Scripts/Services/RestClientService.ts
@@ -580,7 +580,6 @@ module Sfx {
                 + replicaId + "/$/"
                 + "Delete";
             return this.post(this.getApiUrl(url), "Primary Replica Node Move", null, messageHandler);
-
         }
 
         private getEvents<T extends FabricEventBase>(eventType: new () => T, url: string, startTime?: Date, endTime?: Date, messageHandler?: IResponseMessageHandler): angular.IPromise<T[]> {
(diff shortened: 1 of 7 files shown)


src/Sfx/App/Scripts/Services/RestClientService.ts
```diff
@@ -571,6 +571,17 @@ module Sfx {
             return this.getEvents(FabricEvent, url, null, null, messageHandler);
         }
 
+        public restartReplica(nodeName: string, partitionId: string, replicaId: string, messageHandler?: IResponseMessageHandler): angular.IPromise<{}> {
+            let url = "Nodes/"
```
tonglingchen: I like following format more because it is easy to get the Url structure 
`/nodes/${nodeName}/$/GetPartitions/${partitionId}/$/GetReplicas/${replicaId}/$/Restart`

src/Sfx/App/Scripts/Services/RestClientService.ts
```diff
@@ -571,6 +571,17 @@ module Sfx {
...
 
+        public restartReplica(nodeName: string, partitionId: string, replicaId: string, messageHandler?: IResponseMessageHandler): angular.IPromise<{}> {
+            let url = "Nodes/"
+                + nodeName + "/$/"
+                + "GetPartitions/"
+                + partitionId + "/$/"
+                + "GetReplicas/"
+                + replicaId + "/$/"
+                + "Restart";
+            return this.post(this.getApiUrl(url), "Restart replica", null, messageHandler);
```
tonglingchen: This API requires api-version=6.0
By default the api-version is 3.0 so I am not sure if this works.

jeffj6123: I set it to 6.0

src/Sfx/App/Scripts/Services/RestClientService.ts
jeffj6123: Fixed to use that.

Dqovi:
thnx

PR #180, opened 2019-04-09
Now you can enter -1 to InstanceCount when creating stateless service.

OlegKarasik:


juho-hanhimaki:
Can we get this feature included? I find it annoying having to create services through powershell when InstanceCount of -1 is needed.

PR #203, opened 2019-07-11
Clipboard

jeffj6123:
Added copy to clipboard to directive for most locations where a name can commonly have periods or slashes.

### DIFF
--- a/src/Sfx-Proxy/proxy.js
+++ b/src/Sfx-Proxy/proxy.js
@@ -9,6 +9,9 @@ const path = require('path');
 let recordRequest = process.argv.includes("-r");
 let replayRequest = process.argv.includes("-p");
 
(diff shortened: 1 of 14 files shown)


src/Sfx-Proxy/proxy.js
```diff
@@ -9,6 +9,9 @@ const path = require('path');
 let recordRequest = process.argv.includes("-r");
 let replayRequest = process.argv.includes("-p");
 
+console.log("record requests : " + recordRequest);
```
Christina-Kang: How noisy will this be?

jeffj6123: it only displays once when you start the script. It felt like a good idea to make sure the flags passed in are correct and made it easier to check when it started up.

PR #204, opened 2019-07-17
advanced mode toggle

jeffj6123:
Added an 'advanced mode' toggle to the settings dropdown that will access additional node actions.
without only advanced mode turned on just "Restart" will display and with it turned on all possibly actions will display.

### DIFF
--- a/src/Sfx/App/Scripts/Models/DataModels/Collections.ts
+++ b/src/Sfx/App/Scripts/Models/DataModels/Collections.ts
@@ -270,6 +270,15 @@ module Sfx {
             return [allNodes, seedNodes].concat(Object.keys(counts).map(key => counts[key]));
         }
 
+        public setAdvancedMode(state: boolean): void {
+            this.collection.forEach( node => {
+                node.removeAdvancedActions();
+                if(state){
+                    node.setAdvancedActions();
+                }
+            })
+        }
+
         protected get indexPropery(): string {
             // node should be indexed by name
             return "name";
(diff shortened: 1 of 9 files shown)


src/Sfx/App/Scripts/Models/DataModels/Collections.ts
```diff
@@ -270,6 +270,15 @@ module Sfx {
             return [allNodes, seedNodes].concat(Object.keys(counts).map(key => counts[key]));
         }
 
+        public setAdvancedMode(state: boolean): void {
+            this.collection.forEach( node => {
```
Christina-Kang: nit: I would consider moving this loop to its own function, so that it becomes clearer that advanced mode is really a collection different things which can be set to advanced, rather than only the nodes. That can be done later too

src/Sfx/App/Scripts/Models/DataModels/Collections.ts
```diff
@@ -270,6 +270,15 @@ module Sfx {
             return [allNodes, seedNodes].concat(Object.keys(counts).map(key => counts[key]));
         }
 
+        public setAdvancedMode(state: boolean): void {
+            this.collection.forEach( node => {
+                node.removeAdvancedActions();
```
Christina-Kang: Change the function name to include node

src/Sfx/App/Scripts/Models/DataModels/Node.ts
```diff
@@ -87,6 +91,25 @@ module Sfx {
...
+                () => this.restart(),
+                () => true,
+                "Confirm Node Restart",
+                `Restart node ${this.name} from the cluster ${this.data.$location.host()}?`,
+                this.name
+            ));
+        }
+
+        public removeAdvancedActions(): void {
+            this.actions.collection = this.actions.collection.filter(action => ['activateNode', 'removeNodeState', 'deactivePauseNode', 'deactiveRestartNode', 'deactiveRemoveNodeData'].indexOf(action.name) === -1)
```
Christina-Kang: nit: shorten line

PR #210, opened 2019-08-06
Changes for Service Fabric Backup Restore in SFX

aagup:
We have added SFX functionality for Backup and Restore. The operation are supported by Backup Restore System Service

- Create, Edit and Delete Backup Policy.

- Enable and Disable Backup for Application, Service and Partition.

- Suspend and Resume Backup for Application, Service and Partition.

- Trigger and track Backup of partition

- Trigger and track Restore for partition

image015
image016
image017
image018
image019
image020
image003
image004
image005
image006
image007
image008
image009
image010
image011
image012
image013
image014


jeffj6123:
I will review this when coming off oncall

src/Sfx/App/Scripts/Controllers/ClusterViewController.ts
```diff
@@ -1,174 +1,333 @@
...
+                    templateUrl: "partials/update-backup-policy-dialog.html",
+                    controller: ActionController,
+                    resolve: {
+                        action: () => this,
+                    }
+                },
+                null);
+            this.retentionPolicyRequired = false;
+            this.date = "";
+            this.weekDay = ["Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday", "Sunday"];
```
jeffj6123: is there a cleaner way to handle this? what is the dayIndexMapping for also?

src/Sfx/App/Scripts/Controllers/SettingsController.ts
```diff
@@ -22,7 +22,7 @@ module Sfx {
 
             this.data.getNodes().then(nodeCollection => {
                 nodeCollection.setAdvancedMode(this.advancedModeState);
-            })
+            });
         }
```
jeffj6123: Were you able to get the actions to be removed when you toggled them without adding anything here? With the current way we are doing advanced actions, you would want to add a method like the one on node collection for doing the same with the other places that can have backups.

PR #217, opened 2019-08-23
Expand all/close all on some tree entities

jeffj6123:
added another way to open children on the following entities in the tree view
under Applications:
- application type
-  application
-  service partition

under Nodes:
- Deployed application
- Deployed Service Package

under System
- Service

expandall


jeffj6123:
@Christina-Kang  removed for application type.

PR #247, opened 2019-12-18
Introduce Ring View

tg123:
image


src/Sfx/package.json
```diff
@@ -45,13 +45,16 @@
...
+    "cytoscape": "^3.11.0",
+    "cytoscape-popper": "^1.0.4",
     "d3": "3.5.17",
     "jquery": "^3.4.0",
     "jquery-ui": "^1.12.1",
     "lodash": "^4.17.15",
     "moment": "^2.22.2",
     "nouislider": "^14.0.2",
     "path": "^0.12.7",
+    "tippy.js": "^2.6.0",
```
jeffj6123: what is tippy for?

tg123: this is to popup a tooltip when clicking on one of those nodes

jeffj6123:
So how does this look when you have higher counts of nodes like 20 or 50?

src/Sfx/App/Scripts/Directives/RingViewDirective.ts
```diff
@@ -0,0 +1,231 @@
...
+    public restrict = "AE";
+    public replace = true;
+    public templateUrl = "partials/ring-view.html";
+    public scope = {
+        nodes: "=",
+        clusterManifest: "=",
+    };
+
+    public link($scope: any, element: JQuery, attributes: any, ctrl: DetailViewPartController) {
+      let makeTippy = function(node, html){
```
jeffj6123: can we type these inputs and make it an arrow operator. (node, html) => instead of function(node, html)

src/Sfx/App/Scripts/Directives/RingViewDirective.ts
```diff
@@ -0,0 +1,231 @@
...
+          html: html,
+          trigger: 'manual',
+          arrow: true,
+          placement: 'bottom',
+          hideOnClick: false,
+          interactive: true
+        } ).tooltips[0];
+      };
+
+      let hideTippy = function(node){
```
jeffj6123: type this.

src/Sfx/App/Scripts/Directives/RingViewDirective.ts
```diff
@@ -0,0 +1,231 @@
...
+        let m = $($scope.clusterManifest.raw.Manifest);
+        m.find("NotificationEndpoint").each((idx, endpoint) => {
+          let type = $(endpoint).parent().parent().attr("Name");
+          let port = $(endpoint).attr("Port");
+          tmap[type]  = port;
+        });
+
+        let candidateEndpoints = [];
+
+        m.find("Node").each((idx, n) => {
```
jeffj6123: can we change this to .map instead of .each.

src/Sfx/App/Scripts/Directives/RingViewDirective.ts
```diff
@@ -0,0 +1,231 @@
...
+          candidateEndpoints.push($(n).attr("IPAddressOrFQDN") + ":" + tmap[$(n).attr("NodeTypeRef")]);
+        })
+
+        // TODO load from config
```
jeffj6123: is this TODO still needed to be done?

src/Sfx/App/Scripts/Directives/RingViewDirective.ts
```diff
@@ -0,0 +1,231 @@
...
+        hideAllTippies();
+      });   
+
+      $(".main-view").scroll(function(e){
+        hideAllTippies();
+      });
+
+      $scope.buildLabel = function(node: any){
+        // return node.node_name + " (" + node.phase + ")";
+        return node.node_id.substring(0, 5) + " (" + node.phase + ")";
```
jeffj6123: can we use a formatted string here.

src/Sfx/App/Scripts/Directives/RingViewDirective.ts
```diff
@@ -0,0 +1,231 @@
...
+      }
+
+      $scope.messageHandler = function (node: any) {
```
jeffj6123: can we add a node interface or if this is a reference to a node class that already exists in the project can we use that.

src/Sfx/App/Scripts/types/cytoscape.ts
```diff
@@ -0,0 +1,2 @@
+
+declare var cytoscape:any;
```
jeffj6123: is there not a types declaration for this for typescript support? There should be a library @types/<something > for adding these.

src/Sfx/App/Scripts/Directives/RingViewDirective.ts
tg123: can we use `any`? it is returned by server side json

jeffj6123: where exactly is this getting returned from? Is there a spec for what the json looks like?

tg123:
50 nodes

image

PR #256, opened 2020-03-12
Expand tree to fit view button

jeffj6123:
Now the tree view will allow the user to scroll horizontally instead of using ellipses cut offs and additionally when the view port for the tree view is less than the container width add a button that will match the width and expand(as seen below)
expand panel
 will become when clicking the button.
expanded

### DIFF
--- a/src/SfxWeb/src/app/modules/tree/tree-view/tree-view.component.scss
+++ b/src/SfxWeb/src/app/modules/tree/tree-view/tree-view.component.scss
@@ -1,5 +1,10 @@
 @import "src/styles.scss";
 
+.tree-wrapper {
+    display: flex;
+    flex-direction: column;
+    height: 100%;
+}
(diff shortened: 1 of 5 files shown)


jeffj6123:
new potential location. 
new location

jeffj6123:
Final solution 
idea2

src/SfxWeb/src/app/modules/tree/tree-view/tree-view.component.scss
```diff
@@ -21,8 +26,8 @@
 }
 
 .tree-body {
-    overflow-x: auto;
-    overflow-y: auto;
+    // overflow-x: auto;
```
Christina-Kang: nit: please remove if no longer needed, or write a comment explaining why they're commented out. thanks!

PR #262, opened 2020-04-03
feature: ClusterTreeService.ts add Deactivation Intent to node name suffix if intent not set to 'invalid'

jagilber:
add deactivation intent to node name suffix if intent not set to 'invalid' to assist in administration and triaging of cluster

a couple of examples:

image

image


jeffj6123:
very slow update, im sorry.
This is a solid addition, one request though. Can you also add it in the SfxWeb section too, that is for the v2 sfx stuff, which will eventually replace the previous version.

PR #324, opened 2020-07-08
ensure event store doesnt localize.

### LINKED ISSUE #321: SFX Beta: Events timeline dates are in Spanish
SFXenEspanol

Hello, 

The events timeline months/days are in Spanish when using US English locale.  Screenshot attached. 

Thanks, 
Nate

jeffj6123:
close
#321

### DIFF
--- a/src/SfxWeb/src/app/modules/event-store/event-store-timeline/event-store-timeline.component.ts
+++ b/src/SfxWeb/src/app/modules/event-store/event-store-timeline/event-store-timeline.component.ts
@@ -44,7 +44,9 @@ export class EventStoreTimelineComponent implements AfterViewInit, OnChanges {
     let items = new DataSet<DataItem>();
   
     // create visualization
-    this._timeline = new Timeline(this.container.nativeElement, items, groups);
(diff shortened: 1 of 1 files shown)


src/SfxWeb/src/app/modules/event-store/event-store-timeline/event-store-timeline.component.ts
```diff
@@ -44,7 +44,9 @@ export class EventStoreTimelineComponent implements AfterViewInit, OnChanges {
     let items = new DataSet<DataItem>();
   
     // create visualization
-    this._timeline = new Timeline(this.container.nativeElement, items, groups);
+    this._timeline = new Timeline(this.container.nativeElement, items, groups, {
+        locale: 'en_US'
```
Christina-Kang, line 48: This is hardcoding it to english? Can't we read the locale of the user?

jeffj6123: We could read the user's locale but it makes for a weird experience if only one table in all of SFX localizes like that, id rather have a point where localization is handled properly for a much wider amount of SFX.

PR #375, opened 2020-08-24
Display more information for pending safety checks

jeffj6123:
It is now possible to get information about the current partitions that a health check is pending on. When viewing a pending safety check for a node disable or cluster upgrade, there is now an option to get information about the current partition being health checked and a way to view it.

upgrade1
safetycheck
Showing deactivated node related info because when deactivating there is a info related to safety checks.
deactivated node


jeffj6123:
Add the same information to the node page when the node is in disabling.

Issue #438, opened 2020-11-20
Allow whitespace before or after unprovision verification input text


WhitWaldo:
**Is your feature request related to a problem? Please describe.**
In either version, there is often a need to manually remove services, applications and types. Doing this prompts a box that has me provide the full ID of whatever I'm looking to remove. In SFX v1, I can usually just copy/paste the value from the text provided above the box and paste this in. If there was any whitespace before or after the value, it passed validation anyway, which simply eased the pain of a necessary validation step.

In SFX v2, if there is any whitespace present, validation fails and I have to step through the input, find the whitespace and remove it before proceeding. I understand and support the reason for this confirmation, but my applications are often GUIDs (as they are multi-tenanted) and I'm absolutely not typing in these values to delete them. As I'm just copy/pasting, and Chrome often "helpfully" grabs the surrounding whitespace, it would be a nice quality of life improvement for the validation to support whitespace on either end of the pasted value as in v1.

**Describe the solution you'd like**
Allow whitespace on either side of the value being validated in the prompt validating removal of a service, application or type.

**Additional context**
As mentioned, v1 allows this whitespace and v2 presently doesn't. Ideally, I'd just appreciate it going back to the mildly less strict approach taken in v1.

Thank you for your consideration!

PR #441, opened 2020-11-22
Implemented #438

WhitWaldo:
Implemented #438 allowing whitespace around the confirmation value so as to make copy/pasting more convenient


WhitWaldo:
The failed test doesn't appear to have anything to do with my modification as I didn't change anything having to do with the larger application.

jeffj6123:
hey, @WhitWaldo that change would be applied to V1, V2 is in the SfxWeb folder. Also yes that test failing is unrelated, its one im fixing that can be overly quick to fail.

jeffj6123 (approved):

Issue #439, opened 2020-11-20
Edit Backup Policy dialog box takes in invalid default chars (****)


ayushi1411:
While updating backup policy, password fields take **** as the default value.

ayushi1411:
Checkboxes have been added for primary and secondary credentials in file share storage for backup policy while creating and updating the backup policy for the user to specify if the credentials are empty. Otherwise providing the credentials will be required on creating or updating backup policy

PR #434, opened 2020-11-09
added checkboxes for empty credentials for file share backup policy

ayushi1411:
https://msazure.visualstudio.com/One/_workitems/edit/7507750
https://github.com/microsoft/service-fabric-explorer/issues/439

The changes are for adding checkboxes to indicate that the credentials for file share storage is empty while creating or updating backup policy. It is set to false by default. Hence if the primary or secondary credentials are not required the checkbox needs to be checked and username & password fields will be disabled.

PR #442, opened 2020-11-23
Fixes for PR https://github.com/microsoft/service-fabric-explorer/pull/434

ayushi1411:
Fixes related to PR https://github.com/microsoft/service-fabric-explorer/pull/434
https://github.com/microsoft/service-fabric-explorer/issues/439


jeffj6123:
What is wrong with the previous pull request?

jeffj6123:
Additionally can we combine this and #443 to have all outstanding BRS changes in one pr.

mevora-msft (approved):

ayushi1411:
> What is wrong with the previous pull request?

Hi Jeffrey, 
In the previous PR, the issue was that when customer changes from file share to azure blob storage and checkboxes are false, it would not save the backup policy even if it is using azure blob since it would still consider the username and password as required field. So when I am changing the storage type to azure, I need to make and not required fields.

PR #443, opened 2020-11-23
REST Api changes to validate storage connection during update and creation of backup policy

ayushi1411:


jeffj6123:
@ayushi1411 should this flag always be set to true? Are there any conditions where it should not be?

ayushi1411:
> @ayushi1411 should this flag always be set to true? Are there any conditions where it should not be?

This flag is to validate the connection to backup storage provided while creating and updating the policy. In SFX we are making it as always validate. If the customer chooses to not validate, they can call it from PS and pass the flag as false.

PR #503, opened 2021-03-23
reword 'Restart' to more precise 'Restart Fabric Services' to minimize confusion on node element

jagilber:
reword 'Restart' to more precise 'Restart Fabric Services' to minimize confusion on node element
- both internally and externally users are confused of the intent of 'Restart' off the node element in SFX.
- without context, and being on a node element, the logical assumption of 'Restart' is that is restarting the node which is incorrect.
- adding additional context of 'Restart Fabric Services' is a minimal change that makes it clear the intent of this action.

- there may need to be a documentation review however to reflect the changes

from:
image

to:
image


v-karbovnichy:
That is so much fun to see AppVeyor is used to run builds on PRs to Azure Service Fabric Explorer.

PR #542, opened 2021-05-11
Azure Managed Identity support for BRS

ayushi1411:


jeffj6123:
@ayushi1411 can you run ng lint --fix=true in the SFXWeb directory.

roshanroutray (approved):

ayushi1411:
> @ayushi1411 can you run ng lint --fix=true in the SFXWeb directory.

@jeffj6123  I have run the above command and it succeeds on command prompt. However it is failing during build here.

PR #558, opened 2021-06-08
Multiple event store

dagzn:
# Event Store modification to use it with multiple elements of SFX
(NOTE) Currently this version of combined-events only works for Cluster and Nodes events.

The main objective of this PR is to add the necessary changes to make the event-store component compatible with working with a collection of elements, instead of the current approach that is fixed for just one element.

The IEventStoreData interface is introduced to collect all the necessary information for each element to be displayed in the event-store. Each instance of the interface contains:
- The EventList associated to its type (Node, Cluster, etc).
- Its own TimelineGenerator.
- An ITimeLineData object to save its information for the timeline.
- A displayName (string) that is used to show its name on the template.

The main changes were applied to the logic of the event-store component. The biggest change was that, since we are working with a collection of elements, we would be working with a collection of ITimeLineData to display and a collection of Observables.
The collection of the timeline data is fixed by creating a merging function to join all of the events and groups in a single instance that is later bound to display. While the Observables issue is solved by using a RxJS operator ( forkJoin ).

For each component of every element in SFX (Application, Node, Replica, etc) the applied changes were the same. What was done is the following:
- The IEventStoreData interface is imported.
- All of the references for the current TimelineGenerator and EventList aredeleted.
- The previously mentioned are replaced by an instance of IEventStoreData.
- This instance is placed in an array that is bound to the event-store component in the template.

## Photos:

sfx1

sfx2

### DIFF
--- a/src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.html
+++ b/src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.html
@@ -21,7 +21,7 @@
         </div>
         <div>
             <div class="btn-group btn-group-toggle" ngbRadioGroup [(ngModel)]="showAllEvents" name="radioBasic">
-                <label class="flat-button checkbox-push solid-grey" ngbButtonLabel *ngIf="timelineGenerator">
+                <label class="flat-button checkbox-push solid-grey" ngbButtonLabel *ngIf="listEventStoreData"><!-- This is no the optimal condition -->
                     <input ngbButton type="radio" [value]="false"> Correlated
                 </label>
                 <label class="flat-button checkbox-push solid-grey" ngbButtonLabel>
@@ -43,8 +43,13 @@
             <app-event-store-timeline [events]="timeLineEventsData" ></app-event-store-timeline>
         </div>
 
-        <app-detail-list [list]="eventsList.collection" [listSettings]="eventsList.settings" [isLoading]="!eventsList.isInitialized"></app-detail-list>
-    
+        <div *ngFor="let eventStoreData of listEventStoreData">
+            <h2> {{eventStoreData.displayName}} </h2>
+            <app-detail-list [list]="eventStoreData.eventsList.collection" 
+            [listSettings]="eventStoreData.eventsList.settings" 
+            [isLoading]="!eventStoreData.eventsList.isInitialized"></app-detail-list>
+        </div>
+        
         <div>
             <div style="float: right; padding: 15px 0 0 0; margin: 0px;">
                 <span class="dropdown-toggle dark-background-link bowtie-icon bowtie-status-error"
--- a/src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.html
+++ b/src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.html
@@ -21,7 +21,7 @@
         </div>
         <div>
             <div class="btn-group btn-group-toggle" ngbRadioGroup [(ngModel)]="showAllEvents" name="radioBasic">
-                <label class="flat-button checkbox-push solid-grey" ngbButtonLabel *ngIf="!checkAllOption()">
+                <label class="flat-button checkbox-push solid-grey" ngbButtonLabel *ngIf="showCorrelatedBtn">
                     <input ngbButton type="radio" [value]="false"> Correlated
                 </label>
                 <label class="flat-button checkbox-push solid-grey" ngbButtonLabel>
--- a/src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.html
+++ b/src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.html
@@ -21,7 +21,7 @@
         </div>
         <div>
             <div class="btn-group btn-group-toggle" ngbRadioGroup [(ngModel)]="showAllEvents" name="radioBasic">
-                <label class="flat-button checkbox-push solid-grey" ngbButtonLabel *ngIf="listEventStoreData"><!-- This is no the optimal condition -->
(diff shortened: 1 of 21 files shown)


src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.ts
```diff
@@ -58,92 +67,136 @@ export class EventStoreComponent implements OnInit, OnDestroy {
...
-          this.resetSelectionProperties();
-          this.isResetEnabled = true;
-          this.eventsList.reload().subscribe( data => {
-              this.setTimelineData();
-          });
-      } else {
-          this.resetSelectionProperties();
+      let eventListChange = false;
+
+      for (const data of this.listEventStoreData) {
```
jeffj6123, line 113: given reload will require a http request, this will be non blocking below.
The code before would set the timeline data after it has waited for the reload function subscription to emit and then set the data. Like below, you should wait until all of them have reloaded and then set the data.

src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.ts
```diff
@@ -58,92 +67,136 @@ export class EventStoreComponent implements OnInit, OnDestroy {
...
+          }
+      }
+
+      if (eventListChange) {
+          this.setTimelineData();
       }
   }
 
+  public mergeTimelineData(combinedData: ITimelineData, data: ITimelineData): void {
+      data.items.forEach(item => {
```
jeffj6123, line 131: remove the { }s to make it a cleaner one line

src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.ts
```diff
@@ -58,92 +67,136 @@ export class EventStoreComponent implements OnInit, OnDestroy {
...
-    });
-    }
-
-    setNewDates(dates: IOnDateChange) {
-        this.debounceHandler.next(dates);
-    }
+      let rawEventlist = [];
+      const combinedTimelineData = this.initializeTimelineData();
+
+      const subs = this.listEventStoreData.map((data) => {
```
jeffj6123, line 156: you can remove the { }s and return to make this a one liner

jeffj6123 (changes requested):

src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.html
```diff
@@ -21,7 +21,7 @@
         </div>
         <div>
             <div class="btn-group btn-group-toggle" ngbRadioGroup [(ngModel)]="showAllEvents" name="radioBasic">
-                <label class="flat-button checkbox-push solid-grey" ngbButtonLabel *ngIf="timelineGenerator">
+                <label class="flat-button checkbox-push solid-grey" ngbButtonLabel *ngIf="!checkAllOption()">
```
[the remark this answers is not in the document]
dagzn: >  For this ngIf since pshowAllEvents is set with checkAllOptions() you can use that isntead.
To make it work I had to create another variable, since if we bind it to pshowAllEvents whenever its value changes the button would disappear.

src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.ts
dagzn: I've added the forkJoin to wait until all observables have finished and also tried to include built in list functions, please let me know your feedback on this part.

src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.ts
```diff
@@ -53,97 +62,124 @@ export class EventStoreComponent implements OnInit, OnDestroy {
...
           this.isResetEnabled = true;
-          this.eventsList.reload().subscribe( data => {
-              this.setTimelineData();
-          });
-      } else {
-          this.resetSelectionProperties();
+          return data.eventsList.reload();
+      });
+
+      if (subs) {
```
jeffj6123, line 114: have this be a length check > 0

src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.ts
```diff
@@ -53,97 +62,124 @@ export class EventStoreComponent implements OnInit, OnDestroy {
...
+      forkJoin(subs).subscribe(() => {
+          for (const data of this.listEventStoreData) {
+              try {
+                  if (this.pshowAllEvents) {
+                      rawEventlist = rawEventlist.concat(data.eventsList.collection.map(event => event.raw));
+
+                  } else if (data.timelineGenerator) {
+                      const d = data.timelineGenerator.generateTimeLineData(data.eventsList.collection.map(event => event.raw), this.startDate, this.endDate);
+
+                      data.timelineData = {
```
jeffj6123, line 152: so this should be directly assignable, looking at it now there was no reason to pick the properties off before. const d doesnt seem necessary and could be directly assigned to the timeline data

src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.ts
```diff
@@ -53,97 +62,117 @@ export class EventStoreComponent implements OnInit, OnDestroy {
...
-        }catch (e) {
-            console.error(e);
-        }
-    });
-    }
-
-    setNewDates(dates: IOnDateChange) {
-        this.debounceHandler.next(dates);
-    }
+      let rawEventlist = [];
```
jeffj6123, line 138: move these into the subscribe callback to keep them as close to their scope as they can.

jeffj6123 (approved):

PR #570, opened 2021-06-17
Handling failed load

dagzn:
## Error handling for loading an event list
Handling the cases where an elements data fails to load. We display the error in the detail-list and also do not show its groups or items in the timeline.

Fixed the issue where we load the data once again each time we switch the tabs Correlated and All. Now the same data is used once again and we only update it when the date range changes.

### DIFF
--- a/src/SfxWeb/src/app/modules/detail-list-templates/detail-list/detail-list.component.html
+++ b/src/SfxWeb/src/app/modules/detail-list-templates/detail-list/detail-list.component.html
@@ -88,11 +88,14 @@
                     </ng-container>
 
                     </ng-container>
-                    <tr *ngIf="isLoading">
+                    <tr *ngIf="isLoading && successfulLoad">
                         <td [attr.colspan]="listSettings.columnSettings.length">Loading...</td>
                     </tr>
-                    <tr *ngIf="sortedFilteredList.length === 0 && !list && !isLoading">
+                    <tr *ngIf="sortedFilteredList.length === 0 && !list && !isLoading && successfulLoad">
                         <td [attr.colspan]="listSettings.columnSettings.length">No items to display.</td>
+                    </tr>
+                    <tr *ngIf="!successfulLoad">
+                        <td [attr.colspan]="listSettings.columnSettings.length">Error: The list of events has failed to load.</td>
                     </tr>
                 </tbody>
             </table>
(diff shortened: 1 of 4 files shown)


src/SfxWeb/src/app/modules/detail-list-templates/detail-list/detail-list.component.html
```diff
@@ -88,12 +88,15 @@
...
-                    <tr *ngIf="isLoading">
+                    <tr *ngIf="isLoading && successfulLoad">
                         <td [attr.colspan]="listSettings.columnSettings.length">Loading...</td>
                     </tr>
-                    <tr *ngIf="sortedFilteredList.length === 0 && !list && !isLoading">
+                    <tr *ngIf="sortedFilteredList.length === 0 && !list && !isLoading && successfulLoad">
                         <td [attr.colspan]="listSettings.columnSettings.length">No items to display.</td>
                     </tr>
+                    <tr *ngIf="!successfulLoad">
+                        <td [attr.colspan]="listSettings.columnSettings.length">Error: The list of events has failed to load.</td>
```
jeffj6123, line 98: change this to a more generic phrase, like

"Items failed to load"

jeffj6123:
So a good thing to do would be show something in the tabs to make it more obvious instead of needing to go to the tab's list

dagzn:
> So a good thing to do would be show something in the tabs to make it more obvious instead of needing to go to the tab's list

Just added some more info regarding that part. Added some icons to show which tabs have an error and displayed a warning stating that some items failed to load. My only concern is that users might think that the error icon is a closing button, let me know what you think about this issue @jeffj6123 

image

src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.html
```diff
@@ -44,18 +44,23 @@
         </div>
         <br>
 
+        <app-health-badge *ngIf="failedRefresh" [text]="' Some items failed to load.'" [showText]="true" badgeClass="badge-warning"></app-health-badge>
```
jeffj6123, line 47: remove the []s around text and showText

src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.html
```diff
@@ -44,18 +44,23 @@
...
+        
         <div *ngIf="listEventStoreData.length > 1; then showTabs else showSingle"></div>
 
         <ng-template #showTabs>
             <div ngbNav #nav="ngbNav">
                 <div ngbNavItem *ngFor="let data of listEventStoreData">
                     <a ngbNavLink class="bar-name">
+                        <app-health-badge *ngIf="!data.eventsList.lastRefreshWasSuccessful"
```
jeffj6123, line 55: remove the []s around text and showText

PR #575, opened 2021-06-28
Updates to Repair Tasks, refresh function and deleting unused variables

dagzn:
### Changes

We update the IEventStoreData interface to its final version, by replacing the utils interface inside it for just the needed function to set the date window.

There are some other minor updates to the event-store component where we try to avoid the usage of the reload function. Also, we change the way we setup initial dates to avoid arbitrarily accessing the first element of the list.

Also we get rid of unused variables in some files, mainly on the Event Store. Some of this changes are due to deleting the timeframe for refreshing the collection, but the majority are unused variables in the event-store component that serve no purpose in this version of SfxWeb.

### Issue

One issue that still is present is the fact that by using the ensureInitialized function. we may receive true values even though we shouldn't (due to a failure in the request to the API). The issue is:

If we get to the event store page without a failure in a request to the API, every collection would have a successful refresh and would be initialized. But, if later on the cluster fails and we can't get info through the API (errors in the requests) we wouldn't be able to catch those errors and show that something has failed. This happens due to the current loginc in the ensureInitialized function, because we would have a previous successful refresh to the collection that is initialized and that would lead to us not refreshing the data and just using  our current collection.

The possible fixes are:

1.  Instead of the ensureInitialized function just use the refresh function. This would mean that we always ask for the updated collection without looking at what happen to the previous requests.
2. Always force a refresh on the ensureInitialized function. This would also fix the issue but this function at the end would behave the exact same as the refresh function
3. Add a period for a forced refresh on the event store. We can set some variables so that every X amount of seconds a flag is raised so that the next call for ensureInitialized on the collections is a forced one. This way we can maintain a combined functionality.

### DIFF
--- a/src/SfxWeb/src/app/Models/DataModels/collections/CollectionBase.ts
+++ b/src/SfxWeb/src/app/Models/DataModels/collections/CollectionBase.ts
@@ -81,6 +81,7 @@ export class DataModelCollectionBase<T extends IDataModel<any>> implements IData
                 return this.update(collection);
             }),
             catchError( err => {
+                this.collection = [];
                 success = false;
                 return of(err);
             })
--- a/src/SfxWeb/src/app/Models/DataModels/collections/CollectionBase.ts
+++ b/src/SfxWeb/src/app/Models/DataModels/collections/CollectionBase.ts
@@ -87,8 +87,8 @@ export class DataModelCollectionBase<T extends IDataModel<any>> implements IData
                 return of(err);
             })
             ).subscribe( () => {
-                if(this.clearOnFailureToUpdate && this.lastRefreshWasSuccessful){
-                    this.collection = [];   
+                if (this.clearOnFailureToUpdate && this.lastRefreshWasSuccessful){
+                    this.collection = [];
                 }
                 this.lastRefreshWasSuccessful = success;
                 this.refreshingPromise.next(success);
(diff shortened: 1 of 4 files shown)


src/SfxWeb/src/app/Models/DataModels/collections/CollectionBase.ts
```diff
@@ -81,6 +81,7 @@ export class DataModelCollectionBase<T extends IDataModel<any>> implements IData
                 return this.update(collection);
             }),
             catchError( err => {
+                this.collection = [];
```
jeffj6123, line 84: im a little hesitant on this. It makes sense for this situation to clear the collection but I am more so thinking in regards to previously how collections are used. I would think we should add a config to collectioBase thats like
clearOnFailureToUpdate and have it default to false and then check here to dictate whether or not to clear the collection.

dagzn: What I noticed was that after a failed request the collection wasn't affected, so the previous information stayed even though it shouldn't. I agree on what you commented, I'll test some things and be sure to add it in a new commit.

src/SfxWeb/src/app/Models/DataModels/collections/CollectionBase.ts
```diff
@@ -76,16 +77,19 @@ export class DataModelCollectionBase<T extends IDataModel<any>> implements IData
...
             this.retrieveNewCollection(messageHandler).pipe(mergeMap(collection => {
                 return this.update(collection);
             }),
             catchError( err => {
-                this.collection = [];
                 success = false;
                 return of(err);
             })
             ).subscribe( () => {
+                if(this.clearOnFailureToUpdate && this.lastRefreshWasSuccessful){
```
dagzn, line 90: We only have to clear the collection if the previous data is valid (last refresh had no problem) and the current one failed. Decided to put it in the subscribe function because it makes more sense to check it once it is done.

PR #576, opened 2021-07-05
Adding remaining stuff to the Event Store

dagzn:
This PR contains the latest changes to the events components + some fixes. The previous accepted changes are now replicated in the remaining modules of SFX, for each of the elements that contains an event store component.

Also a check has been added to make sure that we only add the repair tasks if the repair manager is enabled in the cluster, this is done by checking the cluster's manifest.

### DIFF
--- a/src/SfxWeb/src/app/views/applications/events/events.component.ts
+++ b/src/SfxWeb/src/app/views/applications/events/events.component.ts
@@ -20,7 +20,7 @@ export class EventsComponent extends BaseControllerDirective {
 
    setup() {
     this.listEventStoreData = [
-      this.data.getApplicationEventData(null)
+      this.data.getApplicationEventData()
     ];
    }
 
--- a/src/SfxWeb/src/app/views/applications/events/events.component.ts
+++ b/src/SfxWeb/src/app/views/applications/events/events.component.ts
@@ -19,10 +19,9 @@ export class EventsComponent extends BaseControllerDirective {
    }
 
    setup() {
-    this.listEventStoreData = [{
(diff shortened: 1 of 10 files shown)


src/SfxWeb/src/app/views/applications/events/events.component.ts
```diff
@@ -19,10 +19,9 @@ export class EventsComponent extends BaseControllerDirective {
    }
 
    setup() {
-    this.listEventStoreData = [{
-      eventsList: this.data.createApplicationEventList(null),
-      displayName: 'Applications'
-    }];
+    this.listEventStoreData = [
+      this.data.getApplicationEventData(null)
```
jeffj6123, line 23: nit: remove the null, given its an optional param it would be better to just not put the null.

src/SfxWeb/src/app/views/nodes/events/events.component.ts
```diff
@@ -12,14 +12,20 @@ export class EventsComponent implements OnInit {
...
+  constructor(public data: DataService, public settings: SettingsService) { }
 
   ngOnInit() {
-    this.listEventStoreData = [{
-      eventsList: this.data.createNodeEventList(null),
-      timelineGenerator: new NodeTimelineGenerator(),
-      displayName: 'Nodes'
-    }];
+    this.listEventStoreData = [
+      this.data.getNodeEventData(null)
```
jeffj6123, line 19: nit: remove the null, given its an optional param it would be better to just not put the null.

jeffj6123:
id say fix the 2 nits and then you are good to merge it.

PR #584, opened 2021-07-15
Changes to events page and tests

dagzn:
The changes introduced in this PR are:

- A couple of tests were reorganized and modified to comply with the new functionality of the option-picker. This had to be done to prevent the PR to fail the E2E tests.
- On the timelineGenerator we add some changes to make it optional to group all of the events of one type by the name of the even type (Cluster, Nodes, Applications, etc.) making it useful when you add more events to one timeline. Also there were some changes to two labels to differentiate the names.
- Reorganization of the events page view by putting the option-picker between the timeline and the detail-list. Also leaving the tabs even if we have one element.
- Add a function to the event store component to correctly add or delete an event type.

### DIFF
--- a/src/SfxWeb/cypress/integration/cluster.spec.js
+++ b/src/SfxWeb/cypress/integration/cluster.spec.js
@@ -188,19 +188,24 @@ context('Cluster page', () => {
 
       cy.get('[data-cy=eventtabs]').within(() => {
         cy.contains(CLUSTER_TAB_NAME)
-        cy.contains(REPAIR_TASK_TAB_NAME)
       })
     })
 
     it("repair manager disabled", () => {
+      addRoute('events', 'empty-list.json', apiUrl(`/EventsStore/Cluster/Events?*`))
       addRoute('repair-manager-manifest', 'manifestRepairManagerDisabled.json', manifest_route)
 
       cy.visit('/#/events')
-
-      cy.wait('@getrepair-manager-manifest')
+      cy.wait(['@getevents','@getrepair-manager-manifest'])
 
       cy.get('[data-cy=eventtabs]').within(() => {
         cy.contains(CLUSTER_TAB_NAME)
+      })
+      checkTableErrorMessage(EMPTY_LIST_TEXT);
+
+      cy.get('[sectionName=select-event-types]').click()
+
+      cy.get('[data-cy=option-picker]').within(() => {
         cy.get(REPAIR_TASK_TAB_NAME).should('not.exist')
       })
     })
@@ -212,12 +217,6 @@ context('Cluster page', () => {
         cy.contains(CLUSTER_TAB_NAME)
       })
       checkTableSize(15);
-
-      cy.get('[data-cy=eventtabs]').within(() => {
-        cy.contains(REPAIR_TASK_TAB_NAME).click();
-      })
-
-      checkTableErrorMessage(EMPTY_LIST_TEXT);
     })
 
     it("all events", () => {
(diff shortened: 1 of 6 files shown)


src/SfxWeb/cypress/integration/cluster.spec.js
```diff
@@ -188,19 +188,24 @@ context('Cluster page', () => {
...
-
-      cy.wait('@getrepair-manager-manifest')
+      cy.wait(['@getevents','@getrepair-manager-manifest'])
 
       cy.get('[data-cy=eventtabs]').within(() => {
         cy.contains(CLUSTER_TAB_NAME)
+      })
+      checkTableErrorMessage(EMPTY_LIST_TEXT);
+
+      cy.get('[sectionName=select-event-types]').click()
```
jeffj6123, line 206: given these strings show up multiple times make them a variable at the top of this file.

src/SfxWeb/cypress/integration/cluster.spec.js
```diff
@@ -188,19 +188,24 @@ context('Cluster page', () => {
...
+
+      cy.get('[data-cy=option-picker]').within(() => {
```
jeffj6123, line 208: given these strings show up multiple times make them a variable at the top of this file.

src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.ts
```diff
@@ -177,6 +181,17 @@ export class EventStoreComponent implements OnInit, OnDestroy {
       });
   }
 
+  processData(option: IOptionData){
+    if (option.addToList){
+      this.listEventStoreData.push(option.data);
+    }
+    else{
+      const index = this.listEventStoreData.indexOf(option.data);
```
jeffj6123, line 189: change this to the filter list operation, I think something like this. condense it to 1 line and use the build in list operation.
this.listEventStoreData = this.listEventStoreData.filter(item => item !== option.data)

src/SfxWeb/src/app/Models/eventstore/timelineGenerators.ts
```diff
@@ -186,17 +186,37 @@ export abstract class TimeLineGeneratorBase<T> {
          throw new Error('NotImplementedError');
     }
 
-    generateTimeLineData(events: T[], startOfRange?: Date, endOfRange?: Date): ITimelineData {
+    generateTimeLineData(events: T[], startOfRange?: Date, endOfRange?: Date, nestedGroupLabel?: string): ITimelineData {
         const data = this.consume(events, startOfRange, endOfRange);
         EventStoreUtils.addSubGroups(data.groups);
+        if (nestedGroupLabel){
```
jeffj6123, line 192: add a comment to explain this structure really briefly and what the point of the nested group label is for.

src/SfxWeb/src/app/Models/eventstore/timelineGenerators.ts
```diff
@@ -186,17 +186,37 @@ export abstract class TimeLineGeneratorBase<T> {
...
+            // We should not add the already nested groups to the new event type one.
+            let groupsAlreadyNested: string[] = [];
+            data.groups.forEach(group => {
+                nestedElementGroup.nestedGroups.push(group.id);
+                if (group.nestedGroups){
+                    groupsAlreadyNested = groupsAlreadyNested.concat(group.nestedGroups);
+                }
+            });
+            // If the group is already nested, we remove it from the nested groups of the new one.
+            nestedElementGroup.nestedGroups = nestedElementGroup.nestedGroups.filter(group => groupsAlreadyNested.indexOf(group) === -1);
```
jeffj6123, line 208: use .includes instead of an index check of --1

jeffj6123 (approved):

PR #716, opened 2022-05-01
Minor capitalization fix

WhitWaldo:
Minor correction of "health" to "Health" on partitions detail page


jeffj6123:
thanks for the fix!

PR #724, opened 2022-05-23
fixing approval banner conditions

jeffj6123:


src/SfxWeb/src/app/Models/DataModels/collections/RepairTaskCollection.ts
```diff
@@ -62,7 +61,7 @@ export class RepairTaskCollection extends DataModelCollectionBase<RepairTask> {
         this.longRunningApprovalJob = longRunningApprovalRepairTask;
         this.longestExecutingJob = longRunningExecutingRepairTask;
```
anantshankar17, line 62: Are we using this.longestExecutingJob anywhere ? We probably need a banner for this aswell as now we usually see ICMs for this case. @saketharsh is working on a proposal to cover all these cases. Saket/Jeffrey let's discuss on the next set of updates once this proposal is ready

anantshankar17 (approved):

src/SfxWeb/src/app/Models/DataModels/collections/RepairTaskCollection.ts
jeffj6123: @anantshankar17 we use the longest executing for one of the tiles on the repair jobs page, we could look at adding a banner for it aswell.

PR #730, opened 2022-06-02
Simultaneous App Events for Event

ceelo777:
Added function to fetch ongoing application events for given input application event.

### DIFF
--- a/src/SfxWeb/src/app/Models/eventstore/timelineGenerators.ts
+++ b/src/SfxWeb/src/app/Models/eventstore/timelineGenerators.ts
@@ -551,6 +551,10 @@ export class ApplicationTimelineGenerator extends TimeLineGeneratorBase<Applicat
         // state necessary for some events
         let previousApplicationUpgrade: ApplicationEvent;
         let upgradeApplicationStarted: ApplicationEvent;
+        
+        if (events.length > 0) {
+            let simulEvents = this.getSimultaneousEventsForEvent(events[0], events);
+        }
         const applicationRollBacks: Record<string, {complete: ApplicationEvent, start?: ApplicationEvent}> = {};
         const processExitedGroups: Record<string, DataGroup> = {};
         const containerExitedGroups: Record<string, DataGroup> = {};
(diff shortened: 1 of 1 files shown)


src/SfxWeb/src/app/Models/eventstore/timelineGenerators.ts
```diff
@@ -186,6 +188,45 @@ export abstract class TimeLineGeneratorBase<T> {
          throw new Error('NotImplementedError');
     }
 
+    getSimultaneousEventsForEvent(configs: IConcurrentEventsConfig[], events: DataSet<DataItem>) : IConcurrentEvents[] {
+        /*
```
jeffj6123, line 192: so this should return data a bit differently. we should return only back the events we care about and not all the matching events.

I.E
The objects returned should be for saying a config of ApplicationProcessExit, we should see a list of ApplicationProcessExist events with concurrentEvents list of events that overlap. This is a change in how currently this returns a list of all events as a list that match any of the config's events.

jeffj6123: so it should be more like this.
we care about event X, with causes of Y and Z.

this function should return a list of event Xs where event Xs have a list(relevantEventsType) which are the events happening at the same time that we care about.

src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.ts
```diff
@@ -186,14 +197,97 @@ export class EventStoreComponent implements OnInit, OnDestroy {
...
+            idx++;
+        }
+        let removedEvent = events[idx];
+        events.splice(idx, 1);
+        consumedEvent['kind'] = removedEvent.kind;                  
+        parsedEvents.add(consumedEvent);
+    });
+  }
+
+  private getConcurrentEventsData(): DataSet<DataItem> {
```
jeffj6123, line 222: is this just for testing?

src/SfxWeb/src/app/views/application/events/events.component.ts
```diff
@@ -13,17 +13,25 @@ import { IOptionConfig } from 'src/app/modules/event-store/option-picker/option-
...
   optionsConfig: IOptionConfig;
 
   constructor(protected data: DataService, injector: Injector) {
     super(data, injector);
   }
 
-  setup() {
+  setup() {    
+    // grab event data for all nodes for concurrent events visualization tool    
     this.listEventStoreData = [
```
jeffj6123, line 25: these events should come from inside the timeline viewer for the moment.

jeffj6123 (changes requested):

src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.ts
ceelo777: Yep, just to test sample application events

src/SfxWeb/src/app/Models/eventstore/timelineGenerators.ts
```diff
@@ -186,6 +197,53 @@ export abstract class TimeLineGeneratorBase<T> {
...
+        inputEvents.forEach(inputEvent => {
+            // iterate through all configuration
+            configs.forEach(config => {
+                if (config.eventType == inputEvent.type) {                                        
+                    // iterate through all events to find relevant ones
+                    events.forEach(iterEvent => {
+                        if (!config.relevantEventsType.includes(iterEvent.type)) {
+                            return;
+                        }
+                        if (this.checkOverlappingTime(inputEvent, iterEvent)) inputEvent.related.push(iterEvent);
```
jeffj6123, line 237: use brackets for if statements to make it easier to read.

src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.ts
```diff
@@ -186,14 +200,78 @@ export class EventStoreComponent implements OnInit, OnDestroy {
       }
 
       return combinedTimelineData;
+  }  
+
+  private makeIConcurrentEvent(event: DataItem) : IConcurrentEvents {
```
jeffj6123, line 205: so this might not be necessary. What if instead of making this new object you have IConcurrentEvents inherit(extend) from DataItem, this would let you just pass the dataItems in directly right? You could probably change the reference to IConcurrentEvents to use ITimelineItem  right?

Issue #732, opened 2022-06-06
Incorrect expand/collapse arrows inside the tree view


jdemis:
**Describe the bug**
The dynamic tree menu on the service fabric explorer page of ``v9.0.1017.9590`` has an unexpected arrow direction.

**To Reproduce**
Open SF explorer `v9.0.1017.9590` and evaluate the visual appearance of the tree on the left cluster menu

**Expected behavior**
Arrows should follow the common design practices, e.g. https://material.angular.io/components/tree/overview

**Screenshots**
Current behavior
![image](https://user-images.githubusercontent.com/10913919/172157789-629a0f62-c7d3-42cf-b927-2381c098433d.png)

 -  Service Fabric version: 9.0.1017.9590
 - OS: Windows
 - Browser: Chrome 102.0.5005.63 x64

**Additional context**

PR #733, opened 2022-06-06
Change rotation of the arrow symbol

jdemis:
Fixes https://github.com/microsoft/service-fabric-explorer/issues/732

This PR will make the right arrow when the tree item is collapsed and rotate down when expanded


jeffj6123:
thank you @jdemis  for noticing and submitting a fix!

PR #738, opened 2022-06-17
Adding Node Type pages with relevant information

sidhant012:
image


src/SfxWeb/src/app/services/tree.service.ts
```diff
@@ -124,13 +124,27 @@ export class TreeService {
...
             }));
 
+            const getNodesTypePromise = this.data.getNodes().pipe(map(nodes => {
+                return nodes.nodeTypes.map(node => {
+                    return {
+                        nodeId: IdGenerator.node(node),
+                        displayName: () => node,
+                        selectAction: () => this.routes.navigate(() => nodes.viewTypePath(node)),
+                        childrenQuery: () => this.getNodes(node),
+                        badge: () => nodes.healthState,
```
jeffj6123, line 141: this would likely show an incorrect state right?

if we have 2 node types and in a node is in error in node type A then i think this would also show node type B as having an error.

src/SfxWeb/src/app/app-routing.module.ts
```diff
@@ -41,6 +41,7 @@ const routes: Routes = [
 
   { path: 'node/:nodeName/deployedapp/:appId', loadChildren: () => import(`./views/deployed-application/deployed-application.module`).then(m => m.DeployedApplicationModule) },
   { path: 'node/:nodeName', loadChildren: () => import(`./views/node/node.module`).then(m => m.NodeModule) },
+  { path: 'node-type/:nodeType', loadChildren: () => import(`./views/node-type/node-type.module`).then(m => m.NodeTypeModule)},
```
jeffj6123, line 44: dont use a - use camel case to keep it consistent.

src/SfxWeb/src/app/views/node-type/base/base.component.spec.ts
```diff
@@ -0,0 +1,25 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
```
jeffj6123, line 1: remove this file.

src/SfxWeb/src/app/views/node-type/node-types/node-types.component.ts
```diff
@@ -0,0 +1,69 @@
...
+  tiles: IDashboardViewModel[] = [];
+  typeCollection: any[];
+
+  constructor(protected data: DataService, private settings: SettingsService, injector: Injector) {
+    super(data, injector);
+   }
+
+  setup() {
+    this.nodes = this.data.nodes;
+    this.listSettings = this.settings.getNewOrExistingListSettings('nodes', ['name'], [
```
jeffj6123, line 30: move this to the settings service to reuse it and dont include the node type for the node type page.

src/SfxWeb/src/app/views/node-type/node-types/node-types.component.ts
```diff
@@ -0,0 +1,69 @@
...
+  selector: 'app-node-types',
+  templateUrl: './node-types.component.html',
+  styleUrls: ['./node-types.component.scss']
+})
+export class NodeTypesComponent extends NodeTypeBaseControllerDirective {
+
+  nodes: NodeCollection;
+  listSettings: ListSettings;
+  tiles: IDashboardViewModel[] = [];
+  typeCollection: any[];
```
jeffj6123, line 22: rename this to something like nodes.

PR #740, opened 2022-06-21
make relevant events config objects

ceelo777:


src/SfxWeb/src/app/Models/eventstore/RelatedEventsConfigs.ts
```diff
@@ -1,20 +1,24 @@
...
+                propertyMappings: [["nodeName", "nodeName"]]
+            }
+        ]
+    },
+    {
+        "eventType": "NodeDown",
+        "relevantEventsType": [
+            {
+                eventType: "NodeDeactivateStarted",
+                propertyMappings: [["nodeName", "nodeName"], ["raw.NodeInstance", "raw.NodeInstance"]]
```
jeffj6123, line 20: this might be easier to interpret as a list of objects.

src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.ts
```diff
@@ -19,16 +19,18 @@ export interface IQuickDates {
     hours: number;
 }
 
+export interface IRelevantEventsConfig {
+    eventType: string;
+    propertyMappings: [string, string][];
```
jeffj6123, line 24: this would probably be easier to interpret if it was something like

propertyMappings: {sourceProperty: string, targetProperty: string}[]
this makes it easier to expand later down the line.

sidhant012 (approved):

PR #741, opened 2022-06-23
Repair Task RCA configurations

sidhant012:


src/SfxWeb/src/app/Models/eventstore/timelineGenerators.ts
```diff
@@ -238,19 +240,29 @@ export abstract class TimeLineGeneratorBase<T> {
                         config.relevantEventsType.forEach(relevantEventType => {
 
                             // check if event type is relevant
+                            if(iterEvent instanceof RepairTask) {
```
jeffj6123, line 243: can we move this information onto the repair task event that gets generated from the timeline generator for repair tasks.

jeffj6123 (changes requested):

src/SfxWeb/src/app/Models/eventstore/timelineGenerators.ts
```diff
@@ -227,6 +227,8 @@ export abstract class TimeLineGeneratorBase<T> {
 
         let simulEvents : IConcurrentEvents[] = [];
         let addedEvents : DataItem[] = [];
+        let sourceVal: any;
```
jeffj6123, line 230: is there a reason we moved these variable declarations here?

src/SfxWeb/src/app/Models/eventstore/timelineGenerators.ts
sidhant012: I don't think so. Thats why I was saying that we would need to add kind to the event that originally comes in. This is currently the best way to do it for the POC until we can add more information to events

PR #749, opened 2022-07-06
Added exitReason config and edited visualization

sidhant012:


src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.ts
```diff
@@ -274,13 +281,25 @@ export class EventStoreComponent implements OnInit, OnDestroy {
...
                                         inputEvent.related = [];
                                     }
-                                    if(Utils.result(iterEvent, relevantEventType.result)) {
-                                        action = "Action: " + Utils.result(iterEvent, relevantEventType.result) + "<br/><br/>";
-                                    }
-                                    iterEvent.reasonForEvent = action;
                                     inputEvent.related.push(iterEvent);
                                     addedEvents.push(iterEvent);
                                 }
+                            } else if(relevantEventType.eventType == "self") {
```
jeffj6123, line 287: wouldnt this trigger for every event?

src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.ts
```diff
@@ -274,13 +281,25 @@ export class EventStoreComponent implements OnInit, OnDestroy {
...
+                                let propMaps = true;
```
jeffj6123, line 288: lets move this logic into another function and pass in all that is relevent.

we could then check inside if the source and target are the same and if so set the target to the property, this should cut down on duplicate code and we could look at adding a pass where it does all self reference checks.

src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.ts
```diff
@@ -28,12 +30,13 @@ export interface IPropertyMapping {
...
     propertyMappings: IPropertyMapping[];
-    result: string; //resulting property we want to display for events (ex. Repair Jobs action)
+    action? : string; //mainly being used for self referential purposes
 }
 
 export interface IConcurrentEventsConfig {
     eventType: string; // the event type we are investigating
-    relevantEventsType: IRelevantEventsConfig[]; // possible causes we are considering  
+    relevantEventsType: IRelevantEventsConfig[]; // possible causes we are considering
+    result: string; //resulting property we want to display for events (ex. Repair Jobs action)
```
jeffj6123, line 39: this should be optional?

src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.ts
sidhant012: so should we switch the iterevent and relevantevetntype loops?

src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.ts
sidhant012: I thought we were only making the relevanteventtypes result optional?

src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.ts
sidhant012: there isn't much duplicate code besides for this `if (sourceVal != null && sourceVal != undefined && targetVal != null && targetVal != undefined && sourceVal != targetVal) {
                                        propMaps = false;
                                    }`

PR #754, opened 2022-07-08
Repair task warning for stuck in health check

jeffj6123:
To make it more clear to customers that a repair job that might appear stuck when in restoring, I have added a warning that will point them to investigate the cluster health.

This will check that if there is a timestamp for health check start but not one for health check end  when a repair job is in Preparing or Restoring state for other 10 minutes, to show  a banner
stuckpreparing
.


jeffj6123:
this should check if preparing/restoring safety checks are true in the condition for showing.

saketharsh (approved):

PR #762, opened 2022-07-25
Added Throttle Info for MR Jobs

saketharsh:


saketharsh:
Added new MR Job Throttled Info into GetJobs query to Infrastructure Service.  Updating SFX to read the value, which can help in providing customer information about Job Throttling - 

https://msazure.visualstudio.com/One/_git/WindowsFabric/pullrequest/6344177

anantshankar17 (approved):
@jeffj6123 this change would be presnet only in the 9.1 runtime, so we should be careful while taking the SFX nuget.

jeffj6123:
we can wrap this in a cluster version check

jeffj6123:
I added a function for checking version you can use https://github.com/microsoft/service-fabric-explorer/pull/763

PR #767, opened 2022-07-28
Changes needed for reason parsing engine

sidhant012:


jeffj6123:
I like the idea but there should be a way to specify specifically source and target transforms.

jeffj6123:
this logic should be moved to a utility location. It shouldnt be located on the event store component, so its easier to reuse. Ideally a function which takes in a value and the list of transforms to perform.

sidhant012:
> this logic should be moved to a utility location. It shouldnt be located on the event store component, so its easier to reuse. Ideally a function which takes in a value and the list of transforms to perform.

What do you mean? It is part of the main config and so the logic would be part of the main engine. the transformations occur due to specific order of transformations we specify for each property and with a passed in value for a specific transformation

jeffj6123:
> > this logic should be moved to a utility location. It shouldnt be located on the event store component, so its easier to reuse. Ideally a function which takes in a value and the list of transforms to perform.
> 
> What do you mean? It is part of the main config and so the logic would be part of the main engine. the transformations occur due to specific order of transformations we specify for each property and with a passed in value for a specific transformation

So this logic will be used in 3 places. we will allow for transforming the result action, source and target properties. More so there is a benefit to having it slightly decoupled, i.e anywhere in SFX could use this and this might be useful functionality to expose for extensibility in SFX.

src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.ts
```diff
@@ -271,6 +287,18 @@ export class EventStoreComponent implements OnInit, OnDestroy {
                                 }
                             });
                             if (propMaps) {
+                                if(relevantEventType.selfTransform) {
+                                    let transformations = relevantEventType.selfTransform;
+                                    transformations.forEach(transform => {
```
jeffj6123, line 292: move this into a function to avoid repeating it.

src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.ts
```diff
@@ -323,7 +343,22 @@ export class EventStoreComponent implements OnInit, OnDestroy {
         return simulEvents;
     }
 
-  private getConcurrentEventsData(): DataSet<DataItem> {
+    private getTransformations(transformations: ITransform[], parsed: any) {
```
jeffj6123, line 346: move this to the Transform.ts file to make it reusable.

src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.ts
```diff
@@ -109,6 +117,12 @@ export class EventStoreComponent implements OnInit, OnDestroy {
 
   public simulEvents: IConcurrentEvents[];
 
+  public magicWand: { [K: string]: Function } = {
```
jeffj6123, line 120: move this to the transform.ts, for now we can just have this list initialized in the file. While it would be beneficial to later make it extendable, this isnt something we will do in SFX.

Issue #782, opened 2022-09-29
Sorting Events grid by Timestamp does nothing


abatishchev:
**Describe the bug**

Clicking on the Timestamp column of the application's Events page doesn't sort the grid. Clicking on other columns does.

**To Reproduce**

Go to `https://<url>:19080/Explorer/index.html#/apptype/<app type>/app/<app name>/events`

**Expected behavior**

**Screenshots**

![image](https://user-images.githubusercontent.com/351644/193085880-5b886b17-a2b8-444d-b6c0-2e42a727a6e1.png)

![image](https://user-images.githubusercontent.com/351644/193085961-1cfa5048-fa5d-4580-9d7b-f991b8533b40.png)

 -  Service Fabric version: 9.0.1048.9590
 - OS (for browser not SF): Windows 10
 - Browser/ Browser version; Edge 105.0.1343.50 

**Additional context**

SFRP cluster in Public cloud.

WhitWaldo:
@jeffj6123 This issue can probably be closed following the recent merge.

PR #785, opened 2022-10-18
Fix for #782

WhitWaldo:
I can't figure out how to run a debug build of SFX locally in a manner that allows it to also interact with an already-running SF dev box cluster so it can pull data, but I believe this to be a fix for the timestamp sorting issue in #782 - the string-based path to the property it sorts on had incorrect casing.


jeffj6123:
@WhitWaldo you should be able to modify the config for the sfx-proxy and run that in conjunction with the SfxWeb project.

PR #788, opened 2022-10-28
condense IS which are cross AZ to one IS name to avoid incorrect warnings

jeffj6123:


src/SfxWeb/src/app/Models/DataModels/collections/infrastructureCollection.ts
```diff
@@ -89,7 +89,13 @@ export class InfrastructureCollection extends DataModelCollectionBase<Infrastruc
...
       })
 
-      const nodetypesWithoutEnoughNodes = this.collection.map(is => InfrastructureCollection.stripPrefix(is.name)).filter(is => (counter.entries().find(count => count.key === is)?.value || 0) < 5);
+      //condense cross AZ infrastructure services
+      const condensedIS = new Set<string>();
+      this.collection.forEach(is => {
+        condensedIS.add(InfrastructureCollection.stripPrefix(is.name).split("/")[0]);
+      })
+
+      const nodetypesWithoutEnoughNodes = Array.from(condensedIS).filter(is => (counter.entries().find(count => count.key === is)?.value || 0) < 5);
```
anantshankar17, line 98: @jeffj6123 There are 3 scenarios with IS's:
1. normal IS where the name is like "fabric:/InfrastructureService/nodetypeName"
2. CrossAZ where the IS name is like "fabric:/InfrastructureService/nodetypeName/{guid}"
3. CoordinatedIS where the name is like "fabric:/InfrastructureService/Coordinated_{guid}"

I am guessing for your updates, I assume #1 & #3 will fall under the same category. Is that right ?
Please right test cases/manually validate that these updates work fine for both #1 and #2, by counting the correct set of nodes.
For #3, may be for now we can let it be as is and fall under #1.

anantshankar17 (approved):

src/SfxWeb/src/app/Models/DataModels/collections/infrastructureCollection.ts
jeffj6123: so @anantshankar17 1 this PR now does the following

take each IS in the system service list and strip the fabric:/InfrastructureService/ part first and then split the string on / and take the first item so with the change above we make a set to remove the duplicates generated by cross AZ.  This assumes that a nodetype can not have / which I assume is reasonable?

fabric:/InfrastructureService/nodetypeName -> nodetypeName -> nodetypeName
fabric:/InfrastructureService/nodetypeName/{guid} -> nodetypeName/{guid} -> nodetypeName (this previously was not stripping the guid which is why we have the bug)

Maybe I dont understand how this last one would map to a nodetype exactly? as of right now this would end up with  Coordinated_{guid} trying to match to a nodetype in the cluster with that name, which I assume is not wanted.
fabric:/InfrastructureService/Coordinated_{guid} -> Coordinated_{guid} -> Coordinated_{guid}


Also this brings up the question of should each IS still be a separate tab? Because in this scenario we would have 9 IS tabs? would this be expected and desirable to keep it easy to reason about?

anantshankar17: >>This assumes that a nodetype can not have / which I assume is reasonable?
Right

>>nodetypeName/{guid} -> nodetypeName
This is also fine, we can set the warning for 5 VMs, just for this.

>> should each IS still be a separate tab
Yes, because IS represents a separate FC tenant and have its own MR job handling, independent of others.

The last one, Coordinated, is a more complex scenario and is not yet available for everyone. In this just one IS spans the entire cluster. The only thing we can do there, for when we detect the nodetype name to be Coordinated_{any_guid}, we don't do anything. We already have checks in place for this new scenario, so customers can't misconfigure, so we don't need the warning in the SFX.

PR #789, opened 2022-10-28
simple placement constraints viewer

jeffj6123:
placementConstraintsservice
placementconstraints


src/SfxWeb/src/app/views/node/essentials/essentials.component.scss
```diff
@@ -0,0 +1,8 @@
+.placement-constraint-wrapper {
```
chensation, line 1: Since this class is used in two places, we could place it into Style.scss and selectively overwrite or add some attributes in the component level scss files.

chensation (approved):

PR #803, opened 2022-11-30
Replication view over time

jeffj6123:
Provide a generic solution to deal with periodic events and refactor some of the code for the existing RCA engine to move to using more shared logic.

Currently

replicasOverTime


src/SfxWeb/src/app/Models/eventstore/periodicEventParser.ts
```diff
@@ -0,0 +1,168 @@
...
+        date: property.extendFromStart ? startDate : new Date(states[0].timeStamp),
+        event: states[0]
+      };
+    })
+  }
+
+
+  states.slice(1).forEach(state => {
+    const newValues = getValues(state, property);
+    console.log(newValues)
```
chensation, line 98: console log needed?

src/SfxWeb/src/app/Models/eventstore/periodicEventParser.ts
```diff
@@ -0,0 +1,168 @@
...
+    if(Utils.isDefined(newValues)) {
+      getRemoved(currentValues, newValues).forEach(removedValue => {
+        console.log(removedValue)
```
chensation, line 101: here as well

src/SfxWeb/src/app/Models/eventstore/rcaEngine.ts
```diff
@@ -170,3 +167,22 @@ export const getSimultaneousEventsForEvent = (configs: IConcurrentEventsConfig[]
...
+  const results = configs.map(config => {
+    return {
+      config,
+      events: []
+    }
+  })
+
+  inputEvents.forEach(event => {
+    results.forEach(config => {
+      if(config.config.eventType === event.kind && validMappings(event, event, config.config)) {
```
chensation, line 181: I'm sure there's a good reason for it, but I don't quite understand why validMappings needed to be called with the source and target as the same event

PR #818, opened 2023-01-09
Event store rework

chensation:


src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.ts
```diff
@@ -1,292 +1,191 @@
...
   listSettings?: ListSettings;
   getEvents?(): S[];
   setDateWindow?(startDate: Date, endDate: Date): boolean;
-  timelineResolver?(id: string): boolean; //used to determine if the data contains a given event;
+  objectResolver?(id: string): IDataModel<any>; //used to determine if the data contains a given event;
+}
+
+interface VisReference {
+  name: string,
+  component: Type<any>
```
jeffj6123, line 33: this type should be the  interface we made for visualizations to implement.

src/SfxWeb/src/app/modules/event-store/event-store/event-store.component.ts
```diff
@@ -1,292 +1,191 @@
...
-      .pipe(debounceTime(400), distinctUntilChanged())
-      .subscribe(dates => {
-        this.startDate = new Date(dates.startDate);
-        this.endDate = new Date(dates.endDate);
-        this.setNewDateWindow();
-      });
+  public dateMin: Date;
+
+  private visualizations: VisualizationComponent[] = [];
+  public vizRefs: VisReference[] =
```
jeffj6123, line 57: technically make this an input with a default of this value.

jeffj6123 (commented):
you should try testing this from a local cluster to make sure there are no issues from the proxy.

PR #823, opened 2023-01-28
Dev/t ppesic/federation tab in sfx

pedjaMicrosoft:
**_Federation Visualization in SFX_**

This PR should add another tab in SFX, which should represent basic visualization of federation.


src/SfxWeb/src/app/views/cluster/federation/federation.component.ts
```diff
@@ -0,0 +1,656 @@
...
+    }
+    else {
+      // make something that visually looks like neighborhood
+      angle = 2 * Math.PI / 3 * (index / arr.length + 1.85);
+      node.x = 1000 * Math.cos(angle);
+      node.y = 1000 * Math.sin(angle);
+    }
+    return node;
+  });
+  let container = document.getElementById("mynetwork");
```
chensation, line 106: we shouldn't use document.getElementById in angular. In this case, I think we could use ViewChild() with template reference variables, see: https://stackoverflow.com/questions/39631594/access-template-reference-variables-from-component-class

src/SfxWeb/src/app/views/cluster/federation/federation.component.ts
```diff
@@ -0,0 +1,656 @@
...
+  }
+  network = new Network(container, data, options);
+
+  // this option gives us possibility of "sliding" through cluster
+  // after finished with forming data, correspondingly update buttons and inputs
+  network.on("click", function(params) {
+    if (params.nodes.length == 0) return;
+    let node: NodeData = nodesMap.get(params.nodes[0]);
+    formNeighborhood(node);
+    let rie: HTMLInputElement = document.getElementById("noneAction") as HTMLInputElement;
```
chensation, line 124: we should use data binding if we are changing the value of an input.

src/SfxWeb/src/app/views/cluster/federation/federation.component.ts
```diff
@@ -0,0 +1,656 @@
...
+  ngOnDestroy() {
+    // free the resources
+    this.debouncerHandlerSubscription.unsubscribe();
+    // this.eventStoreHandlerFMSubscription.unsubscribe();
+    // this.eventStoreHandlerSubscription.unsubscribe();
+  }
+
+  // represent selected node's neighborhood
+  onChangeInput(event: any){
+    let hie: HTMLInputElement = document.getElementById("byName") as HTMLInputElement;
```
chensation, line 313: Same for this and all the following, use data binding. See: https://angular.io/guide/two-way-binding, and 
 You can also find some existing examples in this repo, src/app/modules/powershell-commands may have some that could be helpful.

chensation (commented):
Let's remove unused imports and code first.

src/SfxWeb/src/app/views/cluster/federation/federation.component.ts
```diff
@@ -0,0 +1,656 @@
...
+    });
+  }
+
+  // method being called when customer clicks Refresh button in SFX
+  // since Refresh rate can be set to some predefine time (ie, 5 seconds)
+  // this is probably place we don't want to call logics from, especially for big clusters
+  // because that would take a lot oftime (since we are looping through all of them)
+  // therefore, logics implemented should be removed from here and put into ngOnInit() inside subscription
+  // reason for staying in here currently is because of testing needs (timeline bar is not working in mock regime)
+  refresh(messageHandler?: IResponseMessageHandler): Observable<any> {
```
jeffj6123, line 629: refresh should be unnecessary - the expectation should be that the window will only load what is visible within that time range and only query new data when the slider/dates are changed.

src/SfxWeb/src/app/modules/event-store/event-store.module.ts
```diff
@@ -23,6 +23,6 @@ import { ConcurrentEventsVisualizationModule } from '../concurrent-events-visual
     NgbTooltipModule,
     ConcurrentEventsVisualizationModule
   ],
-  exports: [EventStoreComponent, FullDescriptionComponent, RowDisplayComponent, OptionPickerComponent],
+  exports: [EventStoreComponent, FullDescriptionComponent, RowDisplayComponent, OptionPickerComponent, DoubleSliderComponent],
```
jeffj6123, line 26: this import should go away when rebased to use the new event store wrapper component.

jeffj6123 (changes requested):

src/SfxWeb/src/app/views/cluster/federation/federation.component.ts
```diff
@@ -0,0 +1,656 @@
...
+import { NodeEventList, PartitionEventList } from 'src/app/Models/DataModels/collections/Collections';
+import { NodeEvent, PartitionEvent } from 'src/app/Models/eventstore/Events';
+import { IOnDateChange } from 'src/app/modules/event-store/double-slider/double-slider.component';
+import { IEventStoreData, IQuickDates } from 'src/app/modules/event-store/event-store/event-store.component';
+import { IOptionConfig } from 'src/app/modules/event-store/option-picker/option-picker.component';
+import { DataService } from 'src/app/services/data.service';
+import { TelemetryService } from 'src/app/services/telemetry.service';
+import { TimeUtils } from 'src/app/Utils/TimeUtils';
+import { BaseControllerDirective } from 'src/app/ViewModels/BaseController';
+import { Network} from "vis-network/standalone"
```
jeffj6123, line 14: The most significant changed here would be reworking the structure to take advantage of angular data bindings and template bindings to not access elements through getElementById and not set innerHTML/build HTML in the function to set the dom view. These can make the code more complicated and lose the advantage angular provides for keeping the state in the component in sync with the view.

jeffj6123:
make sure you resolve the conflicts and the zip file is not necessary. If there are specific scenarios in focus to show they should be included as E2E tests.

src/SfxWeb/src/app/views/cluster/federation/federation.component.ts
```diff
@@ -0,0 +1,656 @@
...
+  status?: string;
+  phase?: string;  
+  isSeed?: boolean;
+  openFailed?: string;
+  FM?: boolean;
+  FMM?: boolean;
+  leaseStates?: Map<string, string>;
+}
+
+// all global variables and functions below are in needs of network click listener
```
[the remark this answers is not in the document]
chensation: See Jeffery's comment https://github.com/microsoft/service-fabric-explorer/pull/823/files#r1092209436

src/SfxWeb/src/app/views/cluster/federation/federation.component.ts
```diff
@@ -0,0 +1,656 @@
...
+// passing pointers to functions also lose information about "this", thus they are here global
+
+// two hashmaps, since one group of events provides node names and other one node ids
+let nodesMap = new Map<string, NodeData>();
+let nameToIdMap = new Map<string, string>();
+let neighSize: number = 2;
+
+// data shown on node hover
+function formTitle(elem: [string, NodeData]): string {
+  return "Id: " + elem[1].id +
```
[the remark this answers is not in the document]
pedjaMicrosoft: sure, thanks. for these parts that are optionally added - I don't know can I handle it through format string?
I've been searching for some kind of string builder, but seems there is no such thing in typescript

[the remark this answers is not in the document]
pedjaMicrosoft: I've never used sanitizer, so I'm not sure in which SecurityContext to put this in.
when I used SecurityContext.HTML it just replaced every newline char with &#10; and text was unreadable

src/SfxWeb/src/app/views/cluster/newtab/newtab.component.ts
```diff
@@ -0,0 +1,72 @@
...
+HC1(Highcharts);
+HC2(Highcharts);
+HC3(Highcharts);
+
+@Component({
+  selector: 'app-newtab',
+  templateUrl: './newtab.component.html',
+  styleUrls: ['./newtab.component.scss']
+})
+export class NewtabComponent implements OnInit {
```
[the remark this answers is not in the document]
pedjaMicrosoft: for testing yeah, I forgot to delete it

src/SfxWeb/src/app/views/cluster/federation/federation.component.ts
pedjaMicrosoft: yeah, thanks for this

src/SfxWeb/src/app/views/cluster/federation/federation.component.ts
pedjaMicrosoft: I'm not sure if you meant exactly the two way binding, but if you did it didn't turn out very well when I tried to implement it (maybe because I'm doing it bad), especially when I used it to handle radio button logics. in this case, [(ngModel)] has to have special [value] for every button in here which makes whole implementation more difficult and unreadable, in comparison where in current code I just have to handle 3 lines in order for it to work. also, I had big problems with dispatching events since it looks like two way binding has kind of delay and in SFX it looks like the code is bugged. in the repo you've provided I couldn't actually see where this two way is being used, since it is only declared in template view but doesn't appear in .ts file

src/SfxWeb/src/app/views/cluster/cluster.module.ts
```diff
@@ -30,11 +30,16 @@ import { MetricTileComponent } from './metric-tile/metric-tile.component';
 import { InfrastructureViewComponent } from './infrastructure-view/infrastructure-view.component';
 import { InfrastructureJobModule } from 'src/app/modules/infrastructure-job/infrastructure-job.module';
 import { RepairTasksModule } from 'src/app/modules/repair-tasks/repair-tasks.module';
+import { FederationComponent } from './federation/federation.component';
+import { NewtabComponent } from './newtab/newtab.component';
+import { ConcurrentEventsVisualizationModule } from 'src/app/modules/concurrent-events-visualization/concurrent-events-visualization.module';
+import { DoubleSliderComponent } from 'src/app/modules/event-store/double-slider/double-slider.component';
+import { DualDatePickerComponent } from 'src/app/shared/component/dual-date-picker/dual-date-picker.component';
```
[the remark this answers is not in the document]
pedjaMicrosoft: cleaned them up

src/SfxWeb/src/app/views/cluster/federation/federation.component.scss
```diff
@@ -0,0 +1,5 @@
+#mynetwork {
+    width: 1500px;
```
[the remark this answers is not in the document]
pedjaMicrosoft: would now be okay?

src/SfxWeb/src/app/views/cluster/federation/federation.component.ts
```diff
@@ -0,0 +1,656 @@
...
+  
+  if (network !== null) {
+    network.destroy();
+    network = null;
+  }
+  network = new Network(container, data, options);
+
+  // this option gives us possibility of "sliding" through cluster
+  // after finished with forming data, correspondingly update buttons and inputs
+  network.on("click", function(params) {
```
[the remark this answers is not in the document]
pedjaMicrosoft: just wow, thank you SO MUCH for this, I had no idea lambda functions are providing us this "lifehack" in programming

src/SfxWeb/src/app/views/cluster/federation/federation.component.ts
```diff
@@ -0,0 +1,656 @@
...
+import { IEventStoreData, IQuickDates } from 'src/app/modules/event-store/event-store/event-store.component';
+import { IOptionConfig } from 'src/app/modules/event-store/option-picker/option-picker.component';
+import { DataService } from 'src/app/services/data.service';
+import { TelemetryService } from 'src/app/services/telemetry.service';
+import { TimeUtils } from 'src/app/Utils/TimeUtils';
+import { BaseControllerDirective } from 'src/app/ViewModels/BaseController';
+import { Network} from "vis-network/standalone"
+
+// structure that has all the relevant info about particular node
+export interface NodeData {
```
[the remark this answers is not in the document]
pedjaMicrosoft: actually, all of them have to be present all the time, because we're using two maps.
I've added question marks "just in case" but totally unnecessary. thank you for noticing :)

src/SfxWeb/src/app/views/cluster/federation/federation.component.ts
```diff
@@ -0,0 +1,656 @@
...
+    }
+    return node;
+  });
+  let container = document.getElementById("mynetwork");
+  let data = {
+    nodes: nodes,
+    edges: edges,
+  };
+  
+  if (network !== null) {
```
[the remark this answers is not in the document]
pedjaMicrosoft: I've originally used this sequence of code (when calculating node.y in "neighborhood" regime you have to add "+700" in order to see properly)

if (network == null) {
    let container = document.getElementById("mynetwork");
    nodesVis = new DataSet(nodes);
    edgesVis = new DataSet(edges);
    let data = {
      nodes: nodesVis,
      edges: edgesVis,
    };
    network = new Network(container, data, options);
  }
  else {
    nodesVis.clear();
    edgesVis.clear();
    nodesVis.add(nodes);
    edgesVis.add(edges);
  }

but when sliding more than 5 times, it completely bugs out and stops moving (like some buffering is happening and overload after). also in some examples I've seen that current way of dealing with network is applied and is being much faster, so that's the reason for it

src/SfxWeb/src/app/views/cluster/federation/federation.component.ts
```diff
@@ -0,0 +1,656 @@
...
+        hie.dispatchEvent(new Event('change'));
+        break;
+    }
+
+    let str: string = "";
+    for (let node of mapArray) {
+      str += "<option>" + node[1].name + "</option>";
+    }
+    if (str != "") str = "<option>Nodes</option>" + str;
+    document.getElementById('selectActions').innerHTML = str;
```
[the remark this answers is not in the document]
pedjaMicrosoft: yeah definitelly, thanks

src/SfxWeb/src/app/views/cluster/federation/federation.component.ts
pedjaMicrosoft: yes absolutely, I just used it for easier testing, but it's really unnecessary

src/SfxWeb/src/app/modules/event-store/event-store.module.ts
pedjaMicrosoft: I still need to export the new TimePicker component here

src/SfxWeb/src/app/views/cluster/federation/federation.component.ts
pedjaMicrosoft: Thank you very much for detailed review, I updated most of the things so you can take a look and see if there is still something to repair

src/SfxWeb/package.json
```diff
@@ -50,7 +50,7 @@
     "nouislider": "^15.4.0",
     "rxjs": "~6.5.4",
     "tslib": "^2.0.0",
-    "vis-data": "^6.4.1",
```
jeffj6123, line 53: is there a reason vis-data is removed?

src/SfxWeb/src/app/views/cluster/federation/federation.component.ts
jeffj6123: I would say to look into the angular forms for this https://www.positronx.io/angular-radio-buttons-example/ This would also let you have a set up where you know when by subscribing to the value changes event.

src/SfxWeb/src/app/views/cluster/federation/federation.component.scss
```diff
@@ -0,0 +1,5 @@
+#myNetwork {
+    width: 80%;
+    aspect-ratio: 2 / 1;
```
jeffj6123, line 3: better but setting a width of say 50% would likely work better.

src/SfxWeb/src/app/views/cluster/federation/federation.component.ts
```diff
@@ -0,0 +1,612 @@
...
+  // time-picker
+  public dateMin: Date;
+  public startDate: Date;
+  public endDate: Date;
+
+  constructor(protected data: DataService, injector: Injector, private telemService: TelemetryService, private sanitized: DomSanitizer) {
+    super(injector);
+  }
+
+  ngOnInit() {
```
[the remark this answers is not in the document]
pedjaMicrosoft: it calls the ngOnInit from BaseControllerDirective, but I'm not 100% sure what's happening there
(some  injectors are used and we are also subscribing the setup() method in here). when I comment
out that part of the code, Federation in network isn't showing up, so I left it there

src/SfxWeb/package.json
pedjaMicrosoft: yes, it was in conflict with the vis-network library (some incompatiblity between the versions
if I can remember well). firstly, I updated the vis-data to the newest version, but the same
problem occured. then I forced the vis-network install which succeeded, but after that I got some
dependecy issues and version clashes as well. finally, I removed the vis-data and everything worked out

src/SfxWeb/src/app/views/cluster/federation/federation.component.html
```diff
@@ -0,0 +1,47 @@
...
+<label id="labInputNode" for="inputNode"> Search node by: &nbsp; &nbsp; </label>
+<input type="radio" id="byId" name="ctc" checked #byId>
+<label for="byId"> &nbsp; Id &nbsp; </label>
+<input type="radio" id="byName" name="ctc" #byName>
+<label for="byName"> &nbsp; Name &nbsp; &nbsp; </label>
+<input id="inputNode" size="68" (change)="onChangeInput($event)" #inputNode><br/><br/>
+
+<!-- Section for listing and chosing nodes of interest -->
+<label id="labActions" for="actions"> Select action: </label><br>
+<input type="radio" id="noneAction" name="contact" (click)="onClickRadio($event)" #noneAction>
```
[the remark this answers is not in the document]
pedjaMicrosoft: I changed only this group since there can be additional functionalities added,
don't know if you meant every single one in this .html file

Issue #810, opened 2022-12-10
Use UTC time checkbox is inverted


jackbond:
**Describe the bug**
In the "Recent Application and Container Exit Event Reasons" section, if you check the "Use UTC time" box, it uses local time, if you uncheck the box, it uses UTC time.

**To Reproduce**
Check and uncheck the box.

**Expected behavior**
Checking a box labeled "Use UTC time" should result in times being displayed in UTC.

SF version 9.1.1390.9590
Browser: Edge 107.0.1418.62

jeffj6123:
fixed in newest CU

PR #824, opened 2023-01-31
Set events timeline to reflect isUTC flag at initialization

WhitWaldo:
`viz-timeline` package seems to default to UTC if otherwise unspecified, but `event-store-timeline.component.ts` defaults to `isUTC 
 = false`. While these eventually align after the user toggles the UTC checkbox (`isUTC = true` now), it means the timeline shifts back to UTC when the box is again unchecked, which is how it should have started initially before the toggling.

This simply sets the timezone option explicitly at after-view initialization instead of waiting for the user to set it via the checkbox toggle fixing #810


jeffj6123:
improvements taken here as basis for additoinal improvements in pr #829

PR #837, opened 2023-04-04
Accessibility bugs

chensation:
merge fixes for accessibility bugs

### DIFF
--- a/src/SfxWeb/src/app/views/BaseView.ts
+++ b/src/SfxWeb/src/app/views/BaseView.ts
@@ -1,0 +1,5 @@
+import { ElementRef } from "@angular/core";
+
+export interface IBaseView {
+    el: ElementRef;
+}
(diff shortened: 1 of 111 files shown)


src/SfxWeb/src/app/views/BaseView.ts
```diff
@@ -0,0 +1,5 @@
+import { ElementRef } from "@angular/core";
```
jeffj6123, line 1: is this still used?

jeffj6123 (changes requested):

src/SfxWeb/src/app/views/BaseView.ts
chensation: It's not being used, this was originally for the 'skip to content', I asked if we could keep it in case we wanted to use it later.

PR #838, opened 2023-04-05
Naming viewer

jeffj6123:
provide a way to view metrics emitted by the naming service and provide an interface to be able to easily compare how they change over time.
namingviewer


chensation (approved):
looks good, the event store stuff will need to be changed once the chips are implemented, I'll get to that when I'm working on it.

PR #856, opened 2023-05-23
fixed change detection

chensation:
Fixed the delayed change detection bug present in PowerShell commands.  See bug within the attached image.
new commands

### DIFF
--- a/src/SfxWeb/src/app/modules/powershell-commands/command/command.component.ts
+++ b/src/SfxWeb/src/app/modules/powershell-commands/command/command.component.ts
@@ -27,7 +27,7 @@ export class CommandComponent implements OnInit{
     this.requiredParams = this.command.parameters.filter(p => p.required);
     this.optionalParams = this.command.parameters.filter(p => !p.required);
 
-    this.inputForm.valueChanges.subscribe(_ => { this.updateInvalidInputText(); });
+    this.inputForm.valueChanges.subscribe(_ => { this.updateInputs(); });
   }
 
   goToReference(e: any) {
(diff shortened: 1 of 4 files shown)


src/SfxWeb/src/app/modules/powershell-commands/command/command.component.ts
```diff
@@ -42,12 +43,20 @@ export class CommandComponent implements OnInit{
...
     for (let name in this.requiredInputs?.controls) {
-      if (!this.requiredInputs.controls[name]?.valid)
+      this.command.getParam(name).value = this.requiredInputs.controls[name].value;
+      if (!this.requiredInputs.controls[name]?.valid){
         invalids.push(name);
+      }
     }
     this.invalidInputs = invalids.join(', ');
+
+    for (let name in this.optionalInputs?.controls) {
```
jeffj6123, line 56: when is optionalinputs null?

src/SfxWeb/src/app/modules/powershell-commands/command/command.component.ts
```diff
@@ -42,12 +43,20 @@ export class CommandComponent implements OnInit{
     return this.inputForm.controls['optionalInputs'] as UntypedFormGroup;
   }
 
-  updateInvalidInputText() {
+  updateInputs() {
     let invalids = [];
     for (let name in this.requiredInputs?.controls) {
```
jeffj6123, line 48: when is this null?

jeffj6123 (changes requested):

src/SfxWeb/src/app/modules/powershell-commands/command/command.component.ts
chensation: they were erroring out on page load as inputForm was not set yet, undefined/null was being returned by the getters. Checking again today that doesn't seem to be the case anymore(???). Removed.

src/SfxWeb/src/app/modules/powershell-commands/command/command.component.ts
chensation: Removed.

PR #870, opened 2023-08-31
Dev/yisu/fmql diagnosis

GeoffreyMa:


GeoffreyMa:
Added a warning panel in Clluster/Essential page, when FM partition is in quorum loss status. Also included e2e test and made sure all tests are passed.

src/SfxWeb/src/app/Models/RawDataTypes.ts
```diff
@@ -1,3 +1,4 @@
+import { extend } from 'cypress/types/lodash';
```
chensation, line 1: is this import being used? If not can you remove it?

src/SfxWeb/src/app/views/cluster/essentials/essentials.component.ts
```diff
@@ -91,6 +98,17 @@ export class EssentialsComponent extends BaseControllerDirective {
...
       this.clusterUpgradeProgress.refresh(messageHandler),
+      this.RestClient.getPartitionById("00000000-0000-0000-0000-000000000001", messageHandler).pipe(map((partition) => {
+        this.fmQuorumLossStatus = partition.PartitionStatus;
+        if (this.fmQuorumLossStatus == "InQuorumLoss") {
+          this.fmQuorumLossWarning = "The Failover Manager service is in quorum loss state. " +
+          "Cluster may not be responding. " + 
+          "Service failover, automatic recovery will be blocked. " + 
+          "Health/availability state of services/nodes may not be reflected. "
+        }
+        this.partition = partition;
```
chensation, line 109: this.partition is not being used, remove it

src/SfxWeb/src/app/views/cluster/essentials/essentials.component.ts
```diff
@@ -91,6 +98,17 @@ export class EssentialsComponent extends BaseControllerDirective {
       this.nodes.refresh(messageHandler).pipe(map(() => {this.updateItemInEssentials(); })),
       this.systemApp.refresh(messageHandler).pipe(catchError(err => of(null))),
       this.clusterUpgradeProgress.refresh(messageHandler),
+      this.RestClient.getPartitionById("00000000-0000-0000-0000-000000000001", messageHandler).pipe(map((partition) => {
+        this.fmQuorumLossStatus = partition.PartitionStatus;
+        if (this.fmQuorumLossStatus == "InQuorumLoss") {
```
chensation, line 103: use ===

src/SfxWeb/src/app/views/cluster/essentials/essentials.component.ts
```diff
@@ -91,6 +98,17 @@ export class EssentialsComponent extends BaseControllerDirective {
...
+          this.fmQuorumLossWarning = "The Failover Manager service is in quorum loss state. " +
+          "Cluster may not be responding. " + 
+          "Service failover, automatic recovery will be blocked. " + 
+          "Health/availability state of services/nodes may not be reflected. "
```
chensation, line 107: Why are you using + and multiple strings here? If you just want to break up the string so it doesn't look too long, you can use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals), which allows you to break a string into multiple lines. Also, since this string is not used anywhere else and does not need to have any other values, you can initialize it with this value and remove the if statement.

src/SfxWeb/src/app/views/cluster/essentials/essentials.component.ts
chensation, line 103: use ===

src/SfxWeb/src/app/views/cluster/essentials/essentials.component.ts
chensation, line 109: this.partition is not being used, remove it

src/SfxWeb/src/app/views/cluster/essentials/essentials.component.ts
```diff
@@ -91,6 +98,17 @@ export class EssentialsComponent extends BaseControllerDirective {
       this.nodes.refresh(messageHandler).pipe(map(() => {this.updateItemInEssentials(); })),
       this.systemApp.refresh(messageHandler).pipe(catchError(err => of(null))),
       this.clusterUpgradeProgress.refresh(messageHandler),
+      this.RestClient.getPartitionById("00000000-0000-0000-0000-000000000001", messageHandler).pipe(map((partition) => {
```
chensation, line 101: We can probably turn this getPartitionById into its own function in data.service.ts as getFailoverManagerPartition, or alternatively add a comment explaining that the parititon id 00000000-0000-0000-0000-000000000001 points to FM.

src/SfxWeb/src/app/views/cluster/essentials/essentials.component.ts
```diff
@@ -41,12 +43,15 @@ export class EssentialsComponent extends BaseControllerDirective {
   replicasDashboard: IDashboardViewModel;
   upgradesDashboard: IDashboardViewModel;
   upgradeAppsCount = 0;
+  fmQuorumLossStatus : string;
```
jeffj6123, line 46: change this to a boolean

src/SfxWeb/src/app/views/cluster/essentials/essentials.component.html
```diff
@@ -40,6 +40,10 @@ <h3 class="nested-header heading">Throttled Infrastructure Jobs</h3>
   </app-collapse-container>
 </div>
 
+<div *ngIf="fmQuorumLossStatus === 'InQuorumLoss'" style="margin-bottom: 15px;" data-cy="fmql-warning">
```
jeffj6123, line 43: move this to a constant and check in the controller component.

src/SfxWeb/src/app/views/cluster/essentials/essentials.component.ts
```diff
@@ -91,6 +98,10 @@ export class EssentialsComponent extends BaseControllerDirective {
       this.nodes.refresh(messageHandler).pipe(map(() => {this.updateItemInEssentials(); })),
       this.systemApp.refresh(messageHandler).pipe(catchError(err => of(null))),
       this.clusterUpgradeProgress.refresh(messageHandler),
+      this.RestClient.getPartitionById(this.data.getFailoverManagerPartition(), messageHandler).pipe(map((partition) => {
```
jeffj6123, line 101: given the FM partition ID is a constant, we should set it as a constant.

jeffj6123 (changes requested):

src/SfxWeb/src/app/views/cluster/essentials/essentials.component.ts
GeoffreyMa: image
The data I get from API call is in string type. I can change the variable name to "fmPartitionStatus" if it's causing confusion.

src/SfxWeb/src/app/views/cluster/essentials/essentials.component.ts
GeoffreyMa: Should I create a new export in Common/Constants.ts that includes several service partition Ids?

jeffj6123: Yes that works

src/SfxWeb/src/app/Common/Constants.ts
```diff
@@ -224,6 +224,11 @@ export class TelemetryEventNames {
 }
 
 
+export class SystemServicePartitionId {
```
jeffj6123, line 227: make this plural

src/SfxWeb/src/app/views/cluster/essentials/essentials.component.ts
```diff
@@ -91,6 +99,12 @@ export class EssentialsComponent extends BaseControllerDirective {
       this.nodes.refresh(messageHandler).pipe(map(() => {this.updateItemInEssentials(); })),
       this.systemApp.refresh(messageHandler).pipe(catchError(err => of(null))),
       this.clusterUpgradeProgress.refresh(messageHandler),
+      this.RestClient.getPartitionById(SystemServicePartitionId.FailoverManagerId, messageHandler).pipe(map((partition) => {
+        if (partition.PartitionStatus === 'InQuorumLoss') {
```
jeffj6123, line 103: this should be like
this.fmQuorumLossStatus = partition.PartitionStatus === 'InQuorumLoss'

because as it stands now once its set once it will never clear the warning even if it goes out of quorum loss

src/SfxWeb/src/app/views/cluster/essentials/essentials.component.ts
```diff
@@ -61,6 +67,8 @@ export class EssentialsComponent extends BaseControllerDirective {
 
     this.infraCollection = this.data.infrastructureCollection;
     this.infraSettings = this.settings.getNewOrExistingInfrastructureSettings();
+    this.fmQuorumLossWarning = `The Failover Manager service is in quorum loss state. Cluster may not be responding. 
+    Service failover, automatic recovery will be blocked.Health/availability state of services/nodes may not be reflected.`
```
jeffj6123, line 71: The Failover Manager service is in quorum loss state, which can cause disruptions in the cluster. Operations like <list operations that are effected by this> will likely fail or get stuck. 

Word it more like that to make it clear which operations specifically will fail. Additionally do we have a public TSG for how to investigate FM in quorum loss? We should when providing that there is an issue also have guidance for how to handle it readily available

GeoffreyMa: There is an FM deadlock TSG: https://eng.ms/docs/cloud-ai-platform/azure-core/one-fleet-platform/one-fleet-platform-timmall/service-fabric-service/service-fabric-service/tsgs-and-traces-by-component/failover/tsg/fm-deadlock. I can discuss with Manoj to see if we have other public guidance on that.

GeoffreyMa: The above TSG does not apply to this situation. I didn't find existing source for guidance here. But we do know we can use client API to query the state of FM (FMM can still response to getPartition in this case). So I added a link in the text explaining about this. If needed I can add the link of TSG page that explains what quorum loss means and some basic FM concepts.

jeffj6123: my main point being if we can see this issue and present it to the customer we should be able to give some info on how to deal with it

PR #874, opened 2023-10-30
Adding 'Previous Role' to the reconfiguration text

vpavlovicMSFT:
Currently, if a partition is undergoing reconfiguration, there is no means to determine the primary replica, as the only information available to customers via SFX is the target role of the replica. The introduction of 'Previous Role' will significantly reduce the risk of accidentally terminating the primary replica, a scenario that could potentially lead to data loss.

While a partition is undergoing reconfiguration, in addition to the 'Target Role', 'Previous Role' will also be displayed.

PreviousReplicaRole SFX

[ADO work item](https://msazure.visualstudio.com/One/_workitems/edit/25019512)


src/SfxWeb/src/app/Models/DataModels/Replica.ts
```diff
@@ -73,13 +73,31 @@ export class ReplicaOnPartition extends DataModelBase<IRawReplicaOnPartition> {
...
+        const previousRole: string = this.raw.PreviousReplicaRole ? `Previous Role: ${this.raw.PreviousReplicaRole} | ` : '';
+
+        if (this.parent.raw.PartitionStatus === 'Reconfiguring') {
+            return `Reconfiguring - ${previousRole}Target Role: ${this.raw.ReplicaRole}`;
+        }
+
+        return this.raw.ReplicaRole;
+    }
+
+    public get currentRole(): string {
```
vpavlovicMSFT, line 85: It appears that currentRole() is not used, could be removed since I want to change all reconfiguration texts from 'Reconfiguring - Target Role: X' to 'Reconfiguring - Previous Role: Y | Target Role: X'.

src/SfxWeb/src/app/Models/DataModels/Replica.ts
```diff
@@ -73,13 +73,31 @@ export class ReplicaOnPartition extends DataModelBase<IRawReplicaOnPartition> {
...
         if (this.parent.raw.PartitionStatus === 'Reconfiguring') {
             return `Reconfiguring - Target Role: ${this.raw.ReplicaRole}`;
         }
 
         return this.raw.ReplicaRole;
     }
 
+    public get previousRole(): string {
```
jeffj6123, line 93: Shouldnt this always return previous replica role unless its a new replica?

jeffj6123:
could we also add a new column to the replica list for partition which shows previous replica role? this could be useful in investigation

src/SfxWeb/src/app/Models/DataModels/Replica.ts
```diff
@@ -73,13 +73,31 @@ export class ReplicaOnPartition extends DataModelBase<IRawReplicaOnPartition> {
     }
 
     public get role(): string {
+        const previousRole: string = this.raw.PreviousReplicaRole ? `Previous Role: ${this.raw.PreviousReplicaRole} | ` : '';
+
+        if (this.parent.raw.PartitionStatus === 'Reconfiguring') {
+            return `Reconfiguring - ${previousRole}Target Role: ${this.raw.ReplicaRole}`;
```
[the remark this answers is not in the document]
vpavlovicMSFT: > I believe the same logic is also in DeplolyedReplica, that will also need to be changed. In addition, let's add a unit test for this change in replica.cy.ts and deployedReplica.cy.ts

This doesn't apply to DeployedReplica, since it's using a different API call [Get-ServiceFabricDeployedReplica,
](https://learn.microsoft.com/en-us/powershell/module/servicefabric/get-servicefabricdeployedreplica?view=azureservicefabricps) and this API doesn't include the same `PreviousReplicaRole` field as [Get-ServiceFabricReplica.
](https://learn.microsoft.com/en-us/powershell/module/servicefabric/get-servicefabricreplica?view=azureservicefabricps)

> Let's change this to Reconfiguring - ${this.raw.PreviousReplicaRole} -> ${this.raw.ReplicaRole} to save space, and only show Target Role: ${this.raw.ReplicaRole} when there's no previous role.

I agree on this, thanks for the great suggestion.

vpavlovicMSFT:
> could we also add a new column to the replica list for partition which shows previous replica role? this could be useful in investigation

We can't do that, since the API used for getting previous replica role [Get-ServiceFabricReplica
](https://learn.microsoft.com/en-us/powershell/module/servicefabric/get-servicefabricreplica?view=azureservicefabricps) only returns previous role if the partition is undergoing a reconfiguration.

But we can add `Previous Role` in partition view if the partition is reconfiguring.
Something like this:
image

src/SfxWeb/src/app/Models/DataModels/Replica.ts
vpavlovicMSFT: Could potential compatibility problems arise if the customer uses the most recent SFX with an SF version that predates this addition of the `PreviousReplicaRole` to the [Get-ServiceFabricReplica
](https://learn.microsoft.com/en-us/powershell/module/servicefabric/get-servicefabricreplica?view=azureservicefabricps) API? I'm uncertain about the testing procedure for this scenario, or whether it's necessary at all.

vpavlovicMSFT: If there are no issues with backwards compatibility, then we can completely remove this if check and always return `PreviousReplicaRole`.

src/SfxWeb/src/app/Models/DataModels/Replica.ts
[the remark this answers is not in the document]
vpavlovicMSFT: Instead of `->`, is it possible to use a right arrow symbol, e.g. `➜` (\u279c)?

vpavlovicMSFT:
Latest changes to the UI:
image

Also, I need help with adjusting the CSS for replica-tile component. Tile's height doesn't adjust to the added field... It looks like the relevant styles are in src\SfxWeb\src\app\modules\partition-replication\replica-status-container\\**replica-status-container.component.scss**

jeffj6123:
> Latest changes to the UI: image
> 
> Also, I need help with adjusting the CSS for replica-tile component. Tile's height doesn't automatically adjust to the added field... It looks like the relevant styles are in src\SfxWeb\src\app\modules\partition-replication\replica-status-container\**replica-status-container.component.scss**

I only meant the table at the bottom, like an extra column added next to replica role

jeffj6123:
> > > Latest changes to the UI: image
> > > Also, I need help with adjusting the CSS for replica-tile component. Tile's height doesn't automatically adjust to the added field... It looks like the relevant styles are in src\SfxWeb\src\app\modules\partition-replication\replica-status-container*_replica-status-container.component.scss_*
> > 
> > 
> > I only meant the table at the bottom, like an extra column added next to replica role
> 
> Oh, we can do that, but like I already mentioned, we can only show previous role if a partition is undergoing a reconfig.

my mistake I didnt see that in the previous message.

src/SfxWeb/src/app/Models/DataModels/Replica.ts
jeffj6123: We cant necessarily ensure that. So it is good to have the backwards compat in place.

vpavlovicMSFT:
Screenshot 2024-04-04 180414

vpavlovicMSFT:
@jeffj6123 @chensation After a considerable break, this PR has been revived! 😊 Please take a moment to review the most recent updates.

chensation:
@vpavlovicMSFT , I think my requested changes still stands. The same role logic is inside `DeployedReplica.ts` and we should change it as well unless there's a reason to not change it. 
In addition, although you added a `PreviousReplicaRole` field to the cypress fixtures, `replica.cy.ts` and  `deployedReplica.cy.ts` don't have any tests to check on if roles are displaying properly. I think it's good practice to always update and add tests with every feature we add so we can maintain and increase our test coverage.

@jeffj6123 , I know your original comment said you only wanted a "Previous Role" column at the bottom table, but would you say that adding a "Previous Role" row to the Replicator Status would still make it easier to read?
image

this particular change has been reverted, and now looks like this
image

mssusnjar:
Hi @jeffj6123 and @chensation, we addressed the requested change and adjusted the logic in `DeployedReplica.ts`. We would like to complete this change soon. Can you please review the latest version?

chensation:
@mssusnjar can you add tests into `replica.cy.ts` and `deployedReplica.cy.ts` to test for this change?

mssusnjar:
@chensation tests for these scenarios are added in the recent commit.

PR #898, opened 2024-05-29
adding banner for ManagedIdentityWarning

sidhant012:
When customers are using Managed Identity authentication EventStore events may temporarily be unavailable. So adding a warning banner to inform customers of this so that they are aware of this and know there is nothing to worry about

image


src/SfxWeb/src/app/Models/DataModels/Cluster.ts
```diff
@@ -215,6 +216,20 @@ export class ClusterManifest extends DataModelBase<IRawClusterManifest> {
                 this.isRepairManagerEnabled = true;
             }else if (item.getAttribute('Name') === 'EventStoreService'){
                 this.isEventStoreEnabled = true;
+            } else if (item.getAttribute('Name') === 'Diagnostics') {
+                const params = item.getElementsByTagName('Parameter');
+                const paramsArray = Array.from(params);
+
+                // Function to heck if both "IdentityKind" and "IdentityReference" values are present
+                function containsParameterNames(params: Element[], ...names: string[]): boolean {
```
chensation, line 224: remove the function, or move it outside of the if statement

chensation (changes requested):
where is mi-warning being used? I don't see that being checked in here. It should both display the warning and remove event store.

sidhant012:
I don't think we should remove eventStore because cached events could still show up

chensation (approved):

PR #939, opened 2025-06-13
Add disabled services warning during application upgrade

vpavlovicMSFT:
This pull request introduces functionality to display a warning message when an application is undergoing an upgrade and has disabled services. It includes changes to Cypress tests, application models, UI components, and styling to support this feature.

#### Cypress Tests
* Added three new test cases in `src/SfxWeb/cypress/e2e/app.cy.js` to verify the warning behavior under different conditions: 
  1. When services are disabled and the app is upgrading.
  2. When services are disabled but the app is not upgrading.
  3. When the app is upgrading but no services are disabled.

#### Data Models
* Introduced a `disabledServicesCount` getter in the `Application` class to calculate the number of disabled services.

#### UI Components
* Updated `base.component.html` to display a warning message using the `app-warning` component when the app is upgrading and has disabled services.
* Added a helper method `getDisabledServicesWarning` in `base.component.ts` to dynamically generate the warning message text.

#### Styling
* Enhanced the warning message styling in `warning.component.scss` for better alignment, padding, and visual clarity.

### Supporting Data
* Added fixture files `app-type-upgrading.json` and `services-with-disabled.json` to simulate application and service states during tests.

### DIFF
--- a/src/SfxWeb/src/app/views/application/base/base.component.ts
+++ b/src/SfxWeb/src/app/views/application/base/base.component.ts
@@ -65,6 +65,6 @@ export class BaseComponent extends ApplicationBaseControllerDirective implements
   }
 
   getDisabledServicesWarning(): string {
-    return `This application is currently undergoing an upgrade and has ${this.app?.disabledServicesCount} disabled services. Please note that once the upgrade is complete and the services are re-enabled, they may not function as expected or run with the latest version.`;
+    return `This application is currently undergoing an upgrade and has ${this.app?.disabledServicesCount} disabled services. Please note that once the upgrade is complete and the services are re-enabled, they may not function as expected.`;
   }
 }


src/SfxWeb/src/app/views/application/base/base.component.ts
```diff
@@ -64,4 +64,7 @@ export class BaseComponent extends ApplicationBaseControllerDirective implements
     });
   }
 
+  getDisabledServicesWarning(): string {
+    return `This application is currently undergoing an upgrade and has ${this.app?.disabledServicesCount} disabled services. Please note that once the upgrade is complete and the services are re-enabled, they may not function as expected or run with the latest version.`;
```
vpavlovicMSFT:
```
or run with the latest version
```

I need to double check this part as it might not be true.

vpavlovicMSFT:
@jeffj6123 @chensation Please review the PR.

PR #941, opened 2025-07-06
Addition of new property in BackupPolicy

anmbisht10:
BackupPolicy1
This is the PR which adds two new properties to BackupPolicy

The default value of QuickREcovery is true 
and the default value of compressionStrategy is ZIP 

Cypress test is added and tested


chensation (approved):

anmbisht10:
@microsoft-github-policy-service agree company="Microsoft"

PR #943, opened 2025-07-17
Added orchestration view page and fixed events types filtering

KovacStefan:
Added orchestration view page to cluster pages.
Wrote cypess e2e tests for this page
Fixed event type filtering by renaming the eventTypesFilter to eventsTypesFilter

### DIFF
--- a/src/SfxWeb/src/app/views/cluster/orchestration-view/orchestration-view.component.html
+++ b/src/SfxWeb/src/app/views/cluster/orchestration-view/orchestration-view.component.html
@@ -14,22 +14,22 @@
   <label style="display: block; margin-top: 15px; margin-bottom: 15px;">
     <app-toggle [(state)]="balancingToggle" [color]="balancingEventColor"
       (stateChange)="updateBalancingToggle($event)" data-cy="balancing-toggle"></app-toggle>
-    Enable Balancing Operations
+    Show Balancing Operations
   </label>
   <label style="display: block; margin-top: 15px; margin-bottom: 15px;">
     <app-toggle [(state)]="placementToggle" [color]="placementEventColor"
       (stateChange)="updatePlacementToggle($event)" data-cy="placement-toggle"></app-toggle>
-    Enable Placement Operations
+    Show Placement Operations
   </label>
   <label style="display: block; margin-top: 15px; margin-bottom: 15px;">
     <app-toggle [(state)]="constrainCheckToggle" [color]="constraintCheckEventColor"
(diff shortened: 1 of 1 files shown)


src/SfxWeb/src/app/views/cluster/orchestration-view/orchestration-view.component.html
```diff
@@ -0,0 +1,89 @@
...
+  </label>
+  <label style="display: block; margin-top: 15px; margin-bottom: 15px;">
+    <app-toggle [(state)]="constrainCheckToggle" [color]="constraintCheckEventColor"
+      (stateChange)="updateConstrantCheckToggle($event)" data-cy="constrain-check-toggle"></app-toggle>
+    Enable Constraint Check Operations
+  </label>
+  <label style="display: block; margin-top: 15px; margin-bottom: 15px;">
+    <app-toggle [(state)]="otherToggle" [color]="otherEventColor"
+      (stateChange)="updateOtherToggle($event)" data-cy="other-toggle"></app-toggle>
+    Enable Other (ClientAPI) Operations
```
chensation: can we name these Show ... Operations instead of Enable ... Operations? Enable sounds like you are modifying the cluster in some way instead of just toggling the graph to show/hide info

chensation:
what is the reason for creating a separate tab instead of just including this info into the existing events tab?

KovacStefan:
> Are the operations viewable without interacting with the timeline graph? We have accessibility requirements on that front, hence why we have a table with all the event info in the events tab.

Yes, there are the same events that have always been visible inside the partition events tab

KovacStefan:
> what is the reason for creating a separate tab instead of just including this info into the existing events tab?

These events already exist inside the partition events tab.
The reason behind creating a new tab for viewing the same events is to make it easier for customers and DRIs to investigate and filter the reasons behind PLB reconfigurations. PLB teams receive a lot of incidents asking why some replicas moved, and the hope is that with this change, these questions will decrease. You can read more about it here: 
https://dev.azure.com/msazure/One/_workitems/edit/26307906/ 
and here: 
https://microsoft-my.sharepoint.com/:w:/r/personal/chrpap_microsoft_com/_layouts/15/guestaccess.aspx?e=rJMV8b&share=EZAMJWb4UfJNq2IlxP-XpwsBo1xxpcMpu9PQ4N6v_eCLsQ

chensation (approved):

src/SfxWeb/src/app/views/cluster/orchestration-view/orchestration-view.component.html
KovacStefan: Done

PR #949, opened 2025-08-04
Update SFX with ToBeRemovedReplica Expiration Time

JCurley-MSFT:
With the addition of soft-deleted replicas, whether a replica is soft-deleted should be visible within SFX and if a replica is in a soft-deleted state, its expiration time (the latest time automatic cleanup of a soft-deleted replica will occur while the partition is healthy) should be visible in the essentials view. 

The essentials page of the Replica view should also present data to customers even if refreshing of health fails so a consistent view (including the replica expiration time) can be visible.

### DIFF
--- a/src/SfxWeb/src/app/Models/DataModels/Replica.ts
+++ b/src/SfxWeb/src/app/Models/DataModels/Replica.ts
@@ -75,10 +75,15 @@ export class ReplicaOnPartition extends DataModelBase<IRawReplicaOnPartition> {
 
     public get role(): string {
         const { PartitionStatus } = this.parent.raw;
-        const { PreviousReplicaRole, ReplicaRole } = this.raw;
+        const { PreviousReplicaRole, ReplicaRole, IsStopped } = this.raw;
+
+        if (IsStopped)
+        {
+            return `Stopped: ToBeDeleted`;
+        }
 
         if (PartitionStatus !== 'Reconfiguring') {
-            return ReplicaRole;
+            return `${ReplicaRole}`;
         }
 
         if (!PreviousReplicaRole || PreviousReplicaRole === 'None') {
@@ -114,6 +119,10 @@ export class ReplicaOnPartition extends DataModelBase<IRawReplicaOnPartition> {
 
     public get replicaRoleSortPriority(): number {
         return SortPriorities.ReplicaRolesToSortPriorities[this.raw.ReplicaRole] || 0;
+    }
+
+    public get isStopped(): boolean {
+        return this.raw.IsStopped;
     }
 
     protected retrieveNewData(messageHandler?: IResponseMessageHandler): Observable<any> {
--- a/src/SfxWeb/src/app/Models/DataModels/Replica.ts
+++ b/src/SfxWeb/src/app/Models/DataModels/Replica.ts
@@ -75,9 +75,9 @@ export class ReplicaOnPartition extends DataModelBase<IRawReplicaOnPartition> {
 
     public get role(): string {
         const { PartitionStatus } = this.parent.raw;
-        const { PreviousReplicaRole, ReplicaRole, IsStopped } = this.raw;
+        const { PreviousReplicaRole, ReplicaRole, ReplicaIsStopped } = this.raw;
 
-        if (IsStopped)
+        if (ReplicaIsStopped)
         {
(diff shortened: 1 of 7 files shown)


JCurley-MSFT:
> @JCurley-MSFT please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@microsoft-github-policy-service agree company="Microsoft"

src/SfxWeb/src/app/Models/DataModels/Replica.ts
```diff
@@ -75,10 +75,15 @@ export class ReplicaOnPartition extends DataModelBase<IRawReplicaOnPartition> {
...
+        const { PreviousReplicaRole, ReplicaRole, ReplicaStatus} = this.raw;
+
+        if (ReplicaStatus == 'ToBeRemoved')
+        {
+            return `ToBeRemoved`;
+        }
 
         if (PartitionStatus !== 'Reconfiguring') {
-            return ReplicaRole;
+            return `${ReplicaRole}`;
```
jeffj6123: can remove the strong interpolation formatting

src/SfxWeb/src/app/views/replica/essentials/essentials.component.ts
```diff
@@ -28,19 +28,32 @@ export class EssentialsComponent extends ReplicaBaseControllerDirective {
   }
 
   refresh(messageHandler?: IResponseMessageHandler): Observable<any> {
-    return forkJoin([
-      this.replica.health.refresh(messageHandler),
-      this.replica.detail.refresh(messageHandler).pipe(map(() => {
-        if (!this.isSystem) {
+  const safeHealthRefresh$ = this.replica.health.refresh(messageHandler).pipe(
+    catchError(error => {
```
jeffj6123, line 32: was there a concern around if one of these calls fails? Right now we dont update the view if one is unavailable. I am more thinking we might need a cleaner solution to partially update the state like essential items.

jeffj6123 (changes requested):

src/SfxWeb/src/app/Models/DataModels/Replica.ts
JCurley-MSFT: Removed!

src/SfxWeb/src/app/views/replica/essentials/essentials.component.ts
JCurley-MSFT: Yes, the concern is that if either of these calls fail, the essential items does not get populated resulting in a view like:
image

which prevents the information regarding expiration time for these Down (but in ToBeRemoved) status replicas from portraying their expiration times.

JCurley-MSFT: Do you have any suggestions on what a cleaner solution could look like? Open to opinions on how to improve it.

jeffj6123: oh, could the entry potentially not exist in the health store? One way might be only updating the essential items that are relevant in their respective call.

JCurley-MSFT: Changed to the refresh of detail returns true of false depending on success and if successful will update ProcessID which pulls from replica.details. All other essential items which pulls from replica.raw will always be updated so that if a replica goes down the information is still available (like in the case of a soft deleted replica).

PR #950, opened 2025-08-08
Adding Resources tab to Service view

anazoric0112:
This PR adds visual display for ResourceGovernance metrics for service -> under Resources tab.
Info that is shown in the tab:
- ServicePackage info (name, ServicePackageActivationMode, whether dynamic load reporting is supported based on activatoin mode)
- Cpu metric info (requested amount and limit for CpuCores specified by customer)
- Memory metric info (requested amound and llimit for MemoryInMB specified by customer)

Visual example:
image


src/SfxWeb/cypress/e2e/service.cy.js
```diff
@@ -182,6 +182,22 @@ context('service', () => {
...
+
+        cy.get('[data-cy=navtabs]').within(() => {
+          cy.contains('resources').click();
+        })
+
+        cy.url().should('include', '/resources')
+
+        cy.get("[data-cy=servicepackage-info]").should("exist");
+        cy.get("[data-cy=cpu-info]").should("exist");
+        cy.get("[data-cy=memory-info]").should("exist");
```
chensation, line 201: @anazoric0112  thank you for adding test, please also add check to see if the expected values are displayed in their prospective field, currently this test only checks if the tiles show up. You can modify the fixture files as needed to get the mock http responses to include fields you want.

anazoric0112: @chensation  I finished tests now, you can review the changes. Thanks

PR #959, opened 2025-12-16
Adding SelfReconfiguring Service type in SFX 

PrabhuKGP:
This PR includes changes to display the correct info for 'SelfReconfiguring' service kind, a new type of service apart from Stateful and Stateless.  All the files in cypress folder are E2E tests

Screenshots to show the changes:

**ReplicaRole, State info in Essentials page:**
image

**Reconfiguring state:**
image

Activation state:
image

### DIFF
--- a/src/SfxWeb/cypress/e2e/service.cy.js
+++ b/src/SfxWeb/cypress/e2e/service.cy.js
@@ -363,7 +363,7 @@ context('service', () => {
   })
 
   describe("xss", () => {
-    it.skip("essentials/details", () => {
+    it("essentials/details", () => {
       const alert = `*${windowAlertText}*`;
 
       addDefaultFixtures("xss/");
--- a/src/SfxWeb/src/app/Models/DataModels/Replica.ts
+++ b/src/SfxWeb/src/app/Models/DataModels/Replica.ts
@@ -65,6 +65,10 @@ export class ReplicaOnPartition extends DataModelBase<IRawReplicaOnPartition> {
         return ServiceKindRegexes.Stateless.test(this.raw.ServiceKind);
     }
 
+    public get isSelfReconfiguringService(): boolean {
+        return ServiceKindRegexes.SelfReconfiguring.test(this.raw.ServiceKind);
+    }
+
     public get id(): string {
         return this.raw.ReplicaId || this.raw.InstanceId;
     }
@@ -83,7 +87,7 @@ export class ReplicaOnPartition extends DataModelBase<IRawReplicaOnPartition> {
             return `ToBeRemoved`;
         }
 
-        if (ServiceKindRegexes.SelfReconfiguring.test(this.raw.ServiceKind)) {
+        if (this.isSelfReconfiguringService) {
             if (PartitionStatus !== 'Reconfiguring') {
                 return InstanceRole;
             }
@@ -107,7 +111,7 @@ export class ReplicaOnPartition extends DataModelBase<IRawReplicaOnPartition> {
     }
 
     public get activationState(): string {
-        if (!ServiceKindRegexes.SelfReconfiguring.test(this.raw.ServiceKind)) {
+        if (!this.isSelfReconfiguringService) {
             return 'N/A';
         }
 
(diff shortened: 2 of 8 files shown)


src/SfxWeb/src/app/Models/DataModels/Service.ts
```diff
@@ -66,6 +66,10 @@ export class Service extends DataModelBase<IRawService> {
         return ServiceKindRegexes.Stateless.test(this.raw.ServiceKind);
     }
 
+    public get isSelfReconfiguringService(): boolean {
+        return ServiceKindRegexes.SelfReconfiguring.test(this.raw.ServiceKind);
+    }
+
     public get serviceKindInNumber(): number {
```
chensation, line 73: Do the serviceKindInNumber need to be adjusted? They are used in UpdateServiceDescription

src/SfxWeb/src/app/Models/DataModels/Replica.ts
```diff
@@ -93,6 +106,26 @@ export class ReplicaOnPartition extends DataModelBase<IRawReplicaOnPartition> {
...
+        }
+
+        if (!previousState || previousState === 'None') {
+            return `Reconfiguring - Target State: ${currentState}`;
+        }
+
+        return `Reconfiguring: ${previousState} ${UnicodeConstants.RightArrow} ${currentState}`;
+    }
+
     public get currentRole(): string {
```
chensation: I believe currentRole and previousRole are not currently used. Since getting the role now involves more logic around checking whether or not the service is selfReconfiguring, it would be best to just delete these so we don't mistakenly use them later. Alternatively let me know if this is safe, I don't know how your new raw.InstanceRole  would interact with raw.ReplicaRole

src/SfxWeb/cypress/e2e/service.cy.js
```diff
@@ -333,10 +334,36 @@ context('service', () => {
...
+          cy.contains("Instance Count")
+          cy.contains("Minimum Instance Count")
+        })
+      })
+    })
   })
 
   describe("xss", () => {
-    it("essentials/details", () => {
+    it.skip("essentials/details", () => {
```
chensation: remove the skip

src/SfxWeb/src/app/views/partition/essentials/essentials.component.ts
```diff
@@ -45,6 +45,11 @@ export class EssentialsComponent extends PartitionBaseControllerDirective {
             defaultSortProperties = ['raw.NodeName'];
         }
 
+        // Add Activation State column for Self Reconfiguring Services
+        if (this.partition.isSelfReconfiguringService) {
```
chensation: correct me if I'm wrong, but I don't think any of your tests check to see if this column gets the correct value?

src/SfxWeb/src/app/Models/DataModels/Replica.ts
```diff
@@ -76,12 +76,25 @@ export class ReplicaOnPartition extends DataModelBase<IRawReplicaOnPartition> {
...
+        if (ServiceKindRegexes.SelfReconfiguring.test(this.raw.ServiceKind)) {
+            if (PartitionStatus !== 'Reconfiguring') {
+                return InstanceRole;
+            }
+
+            if(!PreviousSelfReconfiguringInstanceRole || PreviousSelfReconfiguringInstanceRole === 'None') {
+                return `Reconfiguring - Target Role: ${InstanceRole}`;
+            }
+
+            return `Reconfiguring: ${PreviousSelfReconfiguringInstanceRole} ${UnicodeConstants.RightArrow} ${InstanceRole}`;
```
chensation, line 99: Are there any test to check if this shows the correct value?

chensation (changes requested):

src/SfxWeb/src/app/Models/DataModels/Replica.ts
PrabhuKGP: We can either delete it or split it into two functions for Stateful and SelfReconfiguring. 
currentStatefulReplicaRole() and currentSelfReconfiguringInstanceRole()

src/SfxWeb/src/app/Models/DataModels/Replica.ts
PrabhuKGP: Added tests

src/SfxWeb/src/app/views/partition/essentials/essentials.component.ts
PrabhuKGP: Added tests for the same

src/SfxWeb/src/app/Models/DataModels/Replica.ts
PrabhuKGP: Added the new methods

src/SfxWeb/src/app/Models/DataModels/Service.ts
PrabhuKGP: Updated.

PrabhuKGP (commented):
Addressed the comments.

src/SfxWeb/src/app/views/partition/essentials/essentials.component.ts
PrabhuKGP, line 1: The listSettings persists across navigation, so if we visit a self-reconfiguring partition first and navigate to a stateful partition, the old listSettings with the activation state column will still be there. So, removed the check so that it reloads every time

src/SfxWeb/src/app/Models/DataModels/Service.ts
```diff
@@ -222,7 +232,7 @@ export class ServiceType extends DataModelBase<IRawServiceType> {
     }
 
     public get serviceKind(): string {
-        return this.raw.ServiceTypeDescription.IsStateful ? Constants.ServiceKindStateful : Constants.ServiceKindStateless;
+        return this.raw.ServiceTypeDescription.Kind;
     }
 
     public get serviceKindInNumber(): number {
```
chensation, line 238: the change to serviceKind implies IRawServiceType also supports SelfReconfiging. If that's the case this serviceKindInNumber also needs to be changed

chensation (approved):

PR #1022, opened 2026-07-10
Accessbility 5 2026

chensation:


src/Sfx-Proxy/appsettings.json
```diff
@@ -1,6 +1,6 @@
 {
   "TargetCluster": {
-    "Url": "http://localhost:19080"
```
jeffj6123, line 3: dont change default proxy port from 19080 given that is what a local cluster would use

jeffj6123 (approved):

src/Sfx-Proxy/appsettings.json
chensation: accidental commit, can you reapprove?

jeffj6123 (approved):

PR #1033, opened 2026-08-12
[5/5] Post-upgrade fixes: restore interactions + stabilize action dialog

chensation:
**Stacked PR 5 of 5** — base: `pr/4-eslint` (**[4/5]**). Review/merge after **[4/5]**. This is the final slice; merging all five yields the full Angular 22 upgrade.

### Scope
- **Restore interactions after the Angular 22 upgrade** — tree collapse/resize (`app.component`), essential health tile, event-store, imagestore-viewer, manifest, and the create-service command palette.
- **Stabilize action dialog validation** — defer the submit enable/disable via `queueMicrotask` + explicit change detection to fix a change-detection timing issue.

### Verification
- `npm run build:prod` ✅
- `npm run ci:test` (Vitest) ✅ — 97 passing
- `npm run ci:cy-run` (Cypress e2e) ✅ — 171/171 (`service.cy.js` confirmed 19/19)


src/SfxWeb/src/app/modules/action-dialog/action-dialog/action-dialog.component.ts
```diff
@@ -49,6 +50,9 @@ export class ActionDialogComponent implements AfterViewInit {
   }
 
   setSumbitDisable(value: boolean) {
-    this.disableSubmit = value;
+    queueMicrotask(() => {
```
jeffj6123, line 53: why the microtask here?

src/SfxWeb/src/app/modules/imagestore/imagestore-viewer/imagestore-viewer.component.ts
```diff
@@ -33,6 +33,14 @@ export class ImagestoreViewerComponent implements OnInit {
       new ListColumnSettingWithUtcTime('modifiedDate', 'Date modified'),
       new ListColumnSetting('fileCount', 'Count of Files', {sortPropertyPaths: ['isFolder', 'fileCount']})
     ]);
+
+    this.fileListSettings.columnSettings.forEach(column => {
```
jeffj6123, line 37: isnt there an existing pattern for accessing scope from the parent component for the list settings?

src/SfxWeb/src/app/views/application/create-service/create-service.component.html
```diff
@@ -402,7 +402,7 @@ <h1 collapse-header>{{description.isAdvancedOptionCollapsed ? "Show" : "Hide"}}
                         </div>
                         <div class="modal-footer">
                           <button (click)="description.reset()" title="Reset dialog with default settings" class="flat-button">Reset</button>
-                          <button type="submit" class="solid-button blue" (click)="create()" >Create</button>
+                          <button type="submit" class="solid-button blue" (click)="create(createServiceForm)" [disabled]="!createServiceForm.checkValidity()">Create</button>
```
jeffj6123, line 405: why is this passing in the form object instead of referencing it from the component in the function call?

jeffj6123 (changes requested):

src/SfxWeb/src/app/modules/imagestore/imagestore-viewer/imagestore-viewer.component.ts
chensation: getNewOrExistingListSettings caches the ListSettings by name, so on reopen the cached columns still point at the previous ImageStore instance. This loop is trying to resync against the current image store

src/SfxWeb/src/app/modules/action-dialog/action-dialog/action-dialog.component.ts
chensation: the setSubmitDisable is used as callback for the disableSubmit in the child component, which is emitted synchronously in ngOnInit, which runs during the parent's ngAfterViewInit. This is the cause for those ExpressionChangedAfterIsHasBeenCheckeError. Deferring to microtask allows for a clean change detection pass. The same type of fix is used in eventstore as well

src/SfxWeb/src/app/views/application/create-service/create-service.component.html
chensation: fixed

src/SfxWeb/src/app/modules/imagestore/imagestore-viewer/imagestore-viewer.component.ts
jeffj6123: this is definitely still an anti pattern. I think the goal should be move this to wrap in a subscription to pass this in but as it is now its a big awkward updating these refs. id like to see a work item logged to try and move towards much more signal based updates and for computed signals to update properties.

jeffj6123 (approved):

PR #1040, opened 2026-09-14
Migrate auth from ADAL to MSAL (@azure/msal-browser v5)

chensation:
﻿Migrates Service Fabric Explorer web authentication from the deprecated `adal-angular` to `@azure/msal-browser` v5.

## Changes
- Replace `adal-angular` with `@azure/msal-browser` v5 using the redirect flow (`loginRedirect`/`logoutRedirect`/`handleRedirectPromise`); silent token acquisition via the cached refresh token.
- Rename `AdalService` -> `MsalService` and `adal.service.*` -> `msal.service.*`; `redirectUri = origin + pathname` must exactly match the registered SPA reply URL.
- Remove `adal-angular` and `@types/adal-angular`; drop `adal` from `allowedCommonJsDependencies`.
- Add a warning for web redirect auth errors.
image

## Commits
- Migrate auth from ADAL to MSAL (@azure/msal-browser v5)
- add warning for web redirect auth error


src/SfxWeb/src/app/services/msal.service.ts
```diff
@@ -0,0 +1,133 @@
...
+      }
+    } catch (e) {
+      this.authError = this.describeAuthError(e);
+      console.error(e);
+    }
+  }
+
+  // The /token call returns AADSTS9002326 when the reply URL is registered as a "Web" redirect
+  // instead of "Single-page application".
+  private describeAuthError(e: unknown): string {
```
jeffj6123: is there a way to get more context on the errors returned? does MSAL provide well defined error types/codes we can check against? I would rather use their error class to get better typesense.

src/SfxWeb/src/app/services/msal.service.ts
```diff
@@ -0,0 +1,133 @@
...
+  public get isAuthenticated(): boolean {
+    return !!this.userInfo;
+  }
+
+  // Silent-only token acquisition; MSAL renews from the cached refresh token. Interactive
+  // re-auth is not triggered here so concurrent requests can't each fire a full-page redirect.
+  public acquireTokenResilient(resource: string): Observable<string> {
+    const request: SilentRequest = {
+      scopes: [`${resource}/.default`],
+      account: this.userInfo ?? undefined,
```
jeffj6123: userInfo needs to exist right? so why the ?? check

src/SfxWeb/src/app/app-initializers.ts
```diff
@@ -26,9 +26,9 @@ export function initApp(aadService: AdalService, standaloneIntegrationService: S
       await aadService.load().toPromise();
 
       if (aadService.aadEnabled) {
-        aadService.handleWindowCallback();
-        if (!aadService.isAuthenticated) {
-          aadService.login();
+        await aadService.handleWindowCallback();
+        if (!aadService.isAuthenticated && !aadService.authError) {
```
jeffj6123: i feel like on aad errors maybe we should redirect to a distinct page with more context? it might provide a better experience

src/SfxWeb/src/app/app.component.html
```diff
@@ -99,7 +99,9 @@
           }
         </div>
         <div class="view-container" (click)="smallScreenSize ? resize(0): null" (focusin)="smallScreenSize ? resize(0): null">
-
+          @if (msalService.authError) {
```
jeffj6123: should this be a modal? might be easily ignored this way.

jeffj6123 (changes requested):

src/SfxWeb/src/app/services/msal.service.ts
chensation: now explicitly check against the error code. Example response:

{
    "error": "invalid_request",
    "error_description": "AADSTS9002326: Cross-origin token redemption is permitted only for the 'Single-Page Application' client-type. Request origin: 'http://localhost:3000'. Trace ID: c1b3e6c9-24ff-436d-96aa-c00e6fa11000 Correlation ID: 01a09ec2-79b3-73a7-a16b-d3708e3f4e83 Timestamp: 2026-09-14 07:12:18Z",
    "error_codes": [
        9002326
    ],
    "timestamp": "2026-09-14 07:12:18Z",
...

src/SfxWeb/src/app/services/msal.service.ts
chensation: some bad type conversion between null and undefined, cleaned up

src/SfxWeb/src/app/app-initializers.ts
chensation: good point, given none of sfx's ui are function without properly signed in. Converted this to a new page

image

src/SfxWeb/src/app/app.component.html
chensation: same as above

src/SfxWeb/src/app/app-initializers.ts
chensation: image

src/SfxWeb/src/app/services/msal.service.ts
```diff
@@ -0,0 +1,127 @@
...
+    return this.context;
+  }
+
+  public get userInfo(): AccountInfo | undefined {
+    return this.context.getActiveAccount() ?? this.context.getAllAccounts()[0];
+  }
+
+  public get isAuthenticated(): boolean {
+    return !!this.userInfo;
+  }
```
jeffj6123, line 117: This only checks for a cached account, not a valid refresh token. If the cached session is expired/revoked, `isAuthenticated` stays `true`, so `initApp` skips `loginRedirect`, and `acquireTokenResilient` (silent-only) has no interactive fallback — every request would fail until the cache is cleared. Consider treating `InteractionRequiredAuthError` as unauthenticated, or triggering a one-time interactive re-auth when silent acquisition fails.

src/SfxWeb/src/app/services/msal.service.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  // Silent-only token acquisition; MSAL renews from the cached refresh token. Interactive
+  // re-auth is not triggered here so concurrent requests can't each fire a full-page redirect.
+  public acquireTokenResilient(resource: string): Observable<string> {
+    const request: SilentRequest = {
+      scopes: [`${resource}/.default`],
+      account: this.userInfo,
+    };
+
+    return from(this.context.acquireTokenSilent(request).then(result => result.accessToken)).pipe(retry(3));
```
jeffj6123: `this.context.acquireTokenSilent(request)` is invoked eagerly here, before `from(...)` creates the observable, so `.pipe(retry(3))` only resubscribes to the same already-settled Promise and never performs another silent token acquisition. Wrap the call in `defer(() => ...)` so each retry actually re-invokes `acquireTokenSilent`.

src/SfxWeb/src/app/services/msal.service.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+    restClientMock.getAADmetadata = (messageHandler: IResponseMessageHandler): Observable<AadMetadata> => {
+      return of(new AadMetadata({
+        type: 'aad',
+        metadata: {
+          login: 'login',
+          authority: 'auth',
+          client: 'client-id',
+          cluster: 'cluster-id',
+          redirect: 'redirect',
+          tenant: 'tenant-id'
```
jeffj6123: `login: 'login'` combined with `tenant: 'tenant-id'` resolves to authority `"login/tenant-id"`, which is not an absolute URL. `PublicClientApplication` requires an absolute authority URL, so this fixture would throw during construction before reaching the test's assertions. Use a real URL here, as the later tests in this file do (e.g. `'https://login.microsoftonline.com'`).

jeffj6123 (commented):
Re-surfacing three findings from the earlier automated Copilot review that were only listed as "suppressed" summary items (never posted as individually visible inline comments), and so don't appear to have been addressed by the follow-up commits.

src/SfxWeb/src/app/services/msal.service.ts
chensation: fixed with explicit call to acquireTokenRedirect on InteractionRequiredAuthError

src/SfxWeb/src/app/services/msal.service.ts
chensation: fixed with defer

src/SfxWeb/src/app/services/msal.service.spec.ts
chensation: The fixture actually does not throw and passes the test, authority is only validated lazily and the test never actually uses it. However, this has been replaced with actual url to make the test more resilent

jeffj6123 (approved):

PR #1056, opened 2026-09-17
[Node Throttling] Warnings and event views

oscardto:
## Summary

Adds visibility into node message-throttling state in Service Fabric Explorer using the existing EventStore transition events.

## Changes

- Display throttling intervals in Node and Nodes event timelines without replacing existing events.
- Show a warning on the Node page while that node is throttling.
- Show “One or more nodes are throttling” on the Nodes page when applicable.
- Ignore stale throttling events from previous instances after a restart.
- Handle nodes with no throttling events without displaying warnings.

## Testing

- Cypress `node.cy.js` and `nodes.cy.js`: 19/19 passing
- Manually validated active, ended, multiple-node, restart, and no-event scenarios using scripted data.
- Tested local build with a real cluster, behavior is as expected.


src/SfxWeb/src/app/Models/eventstore/timelineGenerators.ts
```diff
@@ -347,6 +347,121 @@ const NodeDeactivateCompleted = 'NodeDeactivateCompleted';
...
+const NodeMessageThrottlingStarted = 'NodeMessageThrottlingStarted';
+const NodeMessageThrottlingEnded = 'NodeMessageThrottlingEnded';
+
+export interface INodeThrottlingIdentity {
+  nodeId: string;
+  instanceId: string;
+  nodeUpAt: string;
+}
+
+export class NodeThrottlingTimelineGenerator extends TimeLineGeneratorBase<NodeEvent> {
```
chensation, line 359: This new generator should not exist. Its generated timeline item should just be part of nodeTimelineGenerator, and its other functionalities should not be part of timelineGenerator, whose intent is for supplying data into timeline for rendering. It is also causing some weird imports, such as data.service.ts import timelineGenerator just for a constant. 

1.generateThrottlingEvent should follow similar precendence as NodeDeactivation.
2. isFromCurrentNodeIncarnation is generic node info, it should go into the node model (DataModels/Node.ts) next to isDeactivating
3. isCurrentlyThrottling, this is consumed by the essential/all-nodes component, and not a timeline concern. Suggest moving it towards NodeCollection

src/SfxWeb/src/app/views/node/essentials/essentials.component.ts
```diff
@@ -107,6 +113,18 @@ export class EssentialsComponent extends NodeBaseControllerDirective {
       this.node.deployedApps.refresh(messageHandler).pipe(map(() => {
         this.deployedApps = this.node.deployedApps;
       })),
+      this.nodeThrottlingEventData.eventsList.refresh(ResponseMessageHandlers.silentResponseMessageHandler).pipe(map(success => {
```
chensation, line 116: guard against isEventStoreEnabled is needed here. This should also not be part of the main forkJoin, which will delay the refresh cycle.

Look into src/SfxWeb/src/app/views/application/essentials/essentials.component.ts for existing example on getting eventstore data using its own subscription

src/SfxWeb/src/app/views/nodes/all-nodes/all-nodes.component.ts
```diff
@@ -42,18 +48,34 @@ export class AllNodesComponent extends BaseControllerDirective {
...
+          this.tiles.push(
+            DashboardViewModel.fromHealthStateCount(type.nodeType, type.nodeType, false, {
+              ErrorCount: type.errorCount,
+              WarningCount: type.warningCount,
+              OkCount: type.okCount
+            })
+          );
+        });
+      })),
+      this.nodeThrottlingEventData.eventsList.refresh(ResponseMessageHandlers.silentResponseMessageHandler)
```
chensation, line 65: same issue as essentials.component.html

src/SfxWeb/src/app/Models/DataModels/collections/Collections.spec.ts
chensation, line 1: Why is a new spec file added here for this test. It should go towards data.service

chensation (changes requested):

src/SfxWeb/src/app/Models/eventstore/timelineGenerators.ts
oscardto: Addressed all three suggestions:

1. Removed the standalone NodeThrottlingTimelineGenerator and integrated its timeline items into NodeTimelineGenerator, following the existing node transition handling.
2. Moved generic instancevalidation to Node.isEventFromCurrentInstance(), next to isDeactivating.
3. Moved current throttling-state calculation to NodeCollection.isCurrentlyThrottling().

DataService now imports the event-kind constants from Events.ts instead of importing a timeline generator. The artificial NodeThrottling event type and factory registration were also removed.

src/SfxWeb/src/app/Models/eventstore/timelineGenerators.ts
```diff
@@ -347,6 +347,121 @@ const NodeDeactivateCompleted = 'NodeDeactivateCompleted';
...
+  }
+
+  consume(events: NodeEvent[], startOfRange: Date, endOfRange: Date): ITimelineData {
+    const items = new DataSet<ITimelineItem>();
+    const nodeEventMap: Record<string, NodeEvent[]> = {};
+    let potentiallyMissingEvents = false;
+
+    events
+      .filter(event => NodeThrottlingTimelineGenerator.eventKinds.includes(event.kind))
+      .forEach(event => {
```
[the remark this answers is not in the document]
oscardto: Addressed. Throttling ranges are now generated directly inside NodeTimelineGenerator.

NodeUp and NodeInstance changes are treated as incarnation boundaries. If a previous incarnation emitted Started without Ended, its interval is capped at the restart boundary and cannot be paired with events from the new incarnation.

A regression test covers this scenario.

src/SfxWeb/src/app/views/node/essentials/essentials.component.ts
oscardto: Addressed. The Node page now waits for the cluster manifest and creates the throttling event query only when EventStore is enabled.

The query runs through an independent subscription outside the main forkJoin, so it does not delay the page refresh cycle. Subscriptions are tracked by the component lifecycle, duplicate initial requests are avoided, and cached warning state is cleared if EventStore becomes disabled.

src/SfxWeb/src/app/views/nodes/all-nodes/all-nodes.component.ts
oscardto: Addressed on the Nodes page using the same pattern as the Node page.

The EventStore request is gated by isEventStoreEnabled, runs independently from the main node refresh, and is tracked by the component lifecycle. Current state is evaluated only after node data is available, duplicate initial requests are prevented, and stale warning state is cleared if EventStore becomes disabled.

src/SfxWeb/src/app/Models/DataModels/collections/Collections.spec.ts
oscardto: Addressed. I removed Collections.spec.ts and moved this coverage to services/data.service.spec.ts, next to the API responsible for creating the throttling event list.

The test verifies both the throttling event filter and the EventStore retention-based query window.

src/SfxWeb/src/app/Models/DataModels/collections/NodeCollection.ts
```diff
@@ -11,6 +11,7 @@ import { HealthStateConstants, NodeStatusConstants, StatusWarningLevel, BannerWa
 import { DataModelCollectionBase } from './CollectionBase';
 import { IDataModel } from '../Base';
 import { RoutesService } from 'src/app/services/routes.service';
+import { NodeEvent, NodeMessageThrottlingEventKinds, NodeMessageThrottlingStarted } from '../../eventstore/Events';
```
chensation, line 14: nit: this does introduce a circular import between data.service, nodeCollection, and Events.ts. It is benign currently as events import of data.service is type only, and gets removed on js build. I'm unsure on this, creating a struct instead to avoid this dependency could be worth it, but it does introduce additional complexity around how the events side would consume said struct

src/SfxWeb/src/app/Models/DataModels/collections/Collections.ts
```diff
@@ -485,8 +485,8 @@ export class ClusterEventList extends EventListBase<ClusterEvent> {
 export class NodeEventList extends EventListBase<NodeEvent> {
     private nodeName?: string;
 
-    public constructor(data: DataService, nodeName?: string) {
-        super(data);
+    public constructor(data: DataService, nodeName?: string, startDate?: Date, endDate?: Date) {
+        super(data, startDate, endDate);
```
chensation, line 489: nit: instead of changing this shared constructor for a singular use case, you could just call setDateWindow similar to what orchestration-view.component.ts is doing

PR #1060, opened 2026-09-19
Inbuild Replica Status

yomahend:
## Summary

Adds a new **"Inbuild Replica Status"** panel to the partition essentials page that shows live build/copy progress for any replica currently building (KVS ESE-backed stores and RC/TStore), instead of only exposing the final `IsInBuild` flag with no visibility into progress.

## What's new

- **`app-replicas-in-build`**: aggregate panel on the partition essentials page, listing every replica the primary reports as in-build (via `ReplicatorStatus.RemoteReplicators`). Each entry collapses to a header row showing the Replica ID and a 5-phase build stepper (Establish Connection → Get Copy Context → Copy → Copy Catchup → Copy Complete).
- **`app-replica-build-progress`**: per-replica detail view, expanded from the panel above:
  - Copy Context sub-phase stepper (own column).
  - Overview section: Copy Context / Copy Type / Copy Sequence details, sourced from ESE-backed KVS copy detail when available.
  - Copy Progress: current/target LSN bar with copy and catchup zones.
  - Info icon linking to `aka.ms/UnderstandingServiceFabricInbuildReplicaStatus`.
- Replicas that report `IsInBuild: true` but have already reached `InbuildPhase: CopyComplete` are filtered out of the list, since that combination lags by one refresh cycle and would otherwise leave a finished replica lingering in the "in build" panel.
- New `RawDataTypes` for `RemoteInbuildReplicaStatus` / `IRawInbuildReplicaCopyDetail` / `IRawKeyValueStoreProviderCopyDetail`, matching the actual `GetDetail` API contract.
- Small shared-component additions to support the above: `IEssentialListItem.infoText` (tooltip icon on essential-item rows), `phase-diagram`'s `fillHeight` input, a focus-visible outline on `collapse-container`'s header button.

## Test coverage

- New Cypress fixtures covering KVS (ESE), RC, and TStore in-build replicas, across all 5 build phases (`replica-detail-inbuild*.json`).
- `partition.cy.js` extended to assert phase-stepper state, greyed/ungreyed sections, and header layout at each phase.

## Other

- `src/Sfx-Proxy/proxy.js`: fixed the dev proxy's catch-all route (`/*` → `/*splat`) — required for Express 5.2.1 (path-to-regexp v6 no longer accepts a bare wildcard). Needed to run the proxy locally while testing this feature; unrelated to the UI change itself.

## Not included

- No changes to the existing per-replica "Replicator Status" section — build progress is surfaced only in the new panel, not duplicated elsewhere.


src/SfxWeb/cypress/e2e/partition.cy.js
```diff
@@ -139,6 +139,231 @@ context('partition', () => {
...
+
+          cy.visit(urlFormatter(appName, serviceName, partitionId));
+          cy.wait(waitRequest);
+
+          cy.get('[data-cy=replicas-in-build]').within(() => {
+            cy.contains('Inbuild Replica Status');
+            cy.get('.in-build-pane').should('have.length', 4);
+          })
+
+          cy.get('[data-cy=build-progress]').scrollIntoView({ offset: { top: -100 } });
```
chensation, line 204: failing test, when multiple in build exist the data-cy selector selects multiple elements, scrollIntoView only accepts one

src/SfxWeb/cypress/e2e/partition.cy.js
```diff
@@ -139,6 +139,231 @@ context('partition', () => {
...
+              // CopyCatchup hasn't been reached yet by this fixture -- no fill, and greyed out
+              cy.get('.lsn-bar-track').eq(1).find('.lsn-bar-fill').should('not.exist');
+              cy.get('.lsn-bar-track').eq(1).should('have.class', 'section-greyed');
+            });
+          })
+
+          // .main-content has its own overflow:auto, so scroll the section itself into
+          // view -- scrollTo('bottom') would jump past it to whatever renders below.
+          cy.get('[data-cy=build-progress]').scrollIntoView({ offset: { top: -100 } });
+          cy.screenshot('replicas-in-build', { capture: 'fullPage' });
```
chensation, line 190: why are we taking a bunch of screenshots in this spec? Cypress will automatically take screenshot for failing tests, while these manual screenshots could be useful during development, we shouldn't leave these in

src/SfxWeb/src/app/modules/partition-replication/replica-build-progress/replica-build-progress.component.ts
```diff
@@ -0,0 +1,315 @@
...
+};
+
+const copyModeReasonInfoText: Record<string, string> = {
+  DefaultPhysicalCopy: 'Full copy with data to send, secondary supports file-stream copy, physical copy is enabled by configuration, and neither probability nor format mismatch forced logical -- physical (file-stream) copy is used.',
+  LogicalCopyProbability: 'Selected via a probability sample against the configured LogicalCopyProbabilityInPercent, which periodically exercises the logical-copy path even when physical would otherwise qualify.',
+  IncompatibleStoreFormatVersion: "Secondary's on-disk store format isn't compatible with the primary's (e.g. mid-upgrade/downgrade across ESE format versions); forces logical copy.",
+  EmptyDatabase: "Primary has no operations to copy; physical file-stream copy isn't attempted.",
+  FileStreamFullCopyNotSupportedBySecondary: "Secondary doesn't support receiving a file-stream full copy.",
+  FullCopyModeConfiguredAsLogical: 'Physical full copy is disabled and the store is explicitly configured with FullCopyMode::Logical.',
+  EnableFileStreamFullCopySetToFalse: 'Physical full copy is disabled via the EnableFileStreamFullCopy config flag (general case).',
```
chensation, line 80: should SFX be the location for these info texts? Normally they would be from runtime instead. Adding them here would make SFX the authority on these replication descriptions when usually SFX is just a downstream consumer, there would also be additional requirements to ensure SFX's description here stays up to date with runtime changes. If they are for some reason exclusive to SFX, they should live inside Model or Constants instead of being declared directly on the component here

src/SfxWeb/src/app/modules/partition-replication/replicas-in-build/replicas-in-build.component.ts
```diff
@@ -0,0 +1,60 @@
...
+  // struct isn't implemented server-side, so it never populates the ESE-shaped fields either.
+  isEseBackedKvs = false;
+
+  sub = new Subscription();
+
+  ngOnChanges(): void {
+    this.primaryReplica = this.replicas?.find(replica => replica.raw.ReplicaRole === 'Primary');
+
+    if (this.primaryReplica) {
+      this.sub.add(this.primaryReplica.detail.refresh().subscribe(() => {
```
chensation, line 36: a sibling component replica-status-container is also calling refresh on primaryReplica, this should just be moved to the parent instead and pass the data down to avoid duplicate calls

src/SfxWeb/src/app/modules/partition-replication/replicas-in-build/replicas-in-build.component.html
```diff
@@ -0,0 +1,23 @@
+@if (inBuildItems.length > 0) {
+  <div class="detail-pane essen-pane" [class.pane-collapsed]="outerCollapsed" data-cy="replicas-in-build">
```
chensation, line 2: moving the primary replica detail out to the parent would also allow this conditional and pane wrappers to reside in the parent component instead, which conforms with the existing style choice

src/SfxWeb/src/app/shared/component/essential-item/essential-item.component.scss
```diff
@@ -1,3 +1,14 @@
+// The value text and its trailing icon (copy/info) are siblings with no gap by default,
+// which makes long values butt right up against the icon glyph -- add breathing room.
+.essential-list-item {
```
chensation, line 3: this is a global touch that affects every essential-item usage when the goal is just to adjust the info icon. Just add a margin to the info icon similar to what was done for the copy icon

src/SfxWeb/src/app/shared/component/collapse-container/collapse-container.component.scss
```diff
@@ -22,4 +22,9 @@ button {
   background-color: transparent;
   border: 0px;
   color: white;
+  outline: none;
+
+  &:focus-visible {
+    outline: 1px solid white;
```
chensation, line 28: this is a global change, can you explain why it was needed?

src/SfxWeb/src/app/modules/partition-replication/replicas-in-build/replicas-in-build.component.ts
```diff
@@ -0,0 +1,60 @@
...
+        const replicaStatus = this.primaryReplica?.detail.raw.ReplicaStatus;
+        this.isEseBackedKvs = replicaStatus?.Kind === 'KeyValueStore' && replicaStatus?.ProviderKind === 'ESE';
+
+        this.inBuildItems = remoteReplicators
+          // IsInBuild lags CopyComplete by one refresh -- exclude it so a replica that
+          // has already finished building doesn't linger in this list.
+          .filter(replicator => replicator.IsInBuild && replicator.RemoteInbuildReplicaStatus?.InbuildPhase !== 'CopyComplete')
+          .map(replicator => ({
+            replicator,
+            replica: this.replicas.find(replica => replica.id === replicator.ReplicaId)
```
chensation, line 47: this gets computed but is never used?

src/SfxWeb/src/app/modules/partition-replication/replicas-in-build/replicas-in-build.component.html
```diff
@@ -0,0 +1,23 @@
...
+    <app-collapse-container sectionName="Inbuild Replica Status" [(collapsed)]="outerCollapsed">
```
chensation, line 3: let's keep this collapsed by default so the entire partition page view isn't taken up by it by default

src/SfxWeb/src/app/modules/partition-replication/replica-build-progress/replica-build-progress.component.html
```diff
@@ -0,0 +1,58 @@
+@if (buildStatus) {
```
chensation, line 1: accessibility issue, the info being displayed isn't wrapping on smaller screens, so they gets cut off. See application upgrade and repair task for the two approach on solving this issue

Image

chensation (changes requested):

